PR #17, opened 2017-09-29
hotfix/response-type-data-type

jasgrg:
The data type for OktaConfig.responseType needs to be extended to allow for either a string or an array of strings.


jmelberg-okta:

PR #18, opened 2017-09-29
Fixed two issues when responseType is id_token only:

jasgrg:
When responseType is set to id_token only there are two issues.
1) handleAuthentication is expecting the tokens variable to be an array, but it is an object. If it is an object then the call to forEach results in an unhandled exception. To prevent this, first check if the tokens is an array and if not create a new array with the tokens as the only entry.
2) if an access_token was not requested, then isAuthenticated will always return false even if a valid id_token is returned. To handle this case, check for either access_token or idToken inside isAuthenticated.


jmelberg-okta:

PR #24, opened 2017-10-12
:seedling: Adds support for sessionToken auth

jmelberg-okta:
- Adds custom-login page support

### DIFF
--- a/packages/okta-angular/README.md
+++ b/packages/okta-angular/README.md
@@ -46,7 +46,7 @@ In your application's `module.ts` file, create a configuration object:
 ```typescript
 // myApp.module.ts
 
-const config = {
+const oktaConfig = {
   issuer: 'https://{yourOktaDomain}.com/oauth2/default',
   redirectUri: 'http://localhost:{port}/implicit/callback',
   clientId: '{clientId}'
@@ -83,9 +83,9 @@ Routes are protected by the `OktaAuthGuard`, which verifies there is a valid `ac
   component: MyProtectedComponent,
   canActivate: [ OktaAuthGuard ]
 }
-``` 
+```
 
-If a user does not have a valid session, they will redirect to the Okta Login Page for authentication. Once authenticated, they will be redirected back to your application's **protected** page.
+If a user does not have a valid session, they will be redirected to the Okta Login Page for authentication. Once authenticated, they will be redirected back to your application's **protected** page.
 
 ### Update your `NgModule`
 Finally, import the `OktaAuthModule` into your `NgModule`, and instantiate it by passing in your configuration object:
@@ -95,10 +95,50 @@ Finally, import the `OktaAuthModule` into your `NgModule`, and instantiate it by
   imports: [
     ...
     RouterModule.forRoot(appRoutes),
-    OktaAuthModule.initAuth(config)
+    OktaAuthModule.initAuth(oktaConfig)
   ],
 })
 export class MyAppModule { }
+```
+
+### Reference
+**Configuration Options**
+  - `issuer` **(required)**: The OpenID Connect `issuer`
+  - `clientId` **(required)**: The OpenID Connect `client_id`
+  - `redirectUri` **(required)**: Where the callback is hosted
+  - `scope` *(optional)*: Reserved or custom claims to be returned in the tokens
+  - `onAuthRequired` *(optional)*: Accepts a callback to make a decision when authentication is required. If not supplied, `okta-angular` will redirect directly to Okta for authentication.
+
+### Using a custom login-page
+The `okta-angular` SDK supports the session token redirect flow for custom login pages. For more information, [see the basic Okta Sign-in Widget functionality](https://github.com/okta/okta-signin-widget#new-oktasigninconfig).
+
+To handle the session-token redirect flow, you can modify the unauthentication callback functionality by adding a `data` attribute directly to your `Route`:
+
+```typescript
+path: 'protected',
+component: MyProtectedComponent,
+canActivate: [ OktaAuthGuard ],
+data: {
+  onAuthRequired: ({oktaAuth, router}) => {
+    // Redirect the user to your custom login page
+    router.navigate(['/custom-login']);
+  }
+}
+```
+
+Alternatively, set this behavior globally by adding it to your configuration object:
+
+```typescript
+const onAuthRequired = ({oktaAuth, router}) => {
+  // Redirect the user to your custom login page
+  router.navigate(['/custom-login']);
+}
+
+const oktaConfig = {
+  issuer: environment.ISSUER,
+  ...
+  onAuthRequired: onAuthRequired
+};
 ```
 
 ## Development
--- a/packages/okta-angular/README.md
+++ b/packages/okta-angular/README.md
@@ -115,25 +115,28 @@ The `okta-angular` SDK supports the session token redirect flow for custom login
 To handle the session-token redirect flow, you can modify the unauthentication callback functionality by adding a `data` attribute directly to your `Route`:
 
 ```typescript
-path: 'protected',
-component: MyProtectedComponent,
-canActivate: [ OktaAuthGuard ],
-data: {
-  onAuthRequired: ({oktaAuth, router}) => {
-    // Redirect the user to your custom login page
-    router.navigate(['/custom-login']);
-  }
+export function onAuthRequired({oktaAuth, router}) {
+  // Redirect the user to your custom login page
+  router.navigate(['/custom-login']);
 }
+
+const appRoutes: Routes = [
+  ...
+  {
+    path: 'protected',
+    component: MyProtectedComponent,
+    canActivate: [ OktaAuthGuard ],
+    data: {
+      onAuthRequired: onAuthRequired
+      }
+    }
+  }
+]
 ```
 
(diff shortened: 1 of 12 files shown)


packages/okta-angular/src/okta/okta.config.ts
```diff
@@ -16,8 +16,9 @@ export interface OktaConfig {
   issuer?: string;
   redirectUri?: string;
   clientId?: string;
-  responseType?: string;
+  responseType?: string | string[];
```
lboyette-okta: I don't think we should allow this. It exposes an implementation detail of auth-js

packages/okta-angular/src/okta/okta.config.ts
```diff
@@ -16,8 +16,9 @@ export interface OktaConfig {
...
   scopes?: any[];
+  onAuthRequired?: any;
```
lboyette-okta: is there a "Function" type?

packages/okta-angular/src/okta/okta.guard.ts
```diff
@@ -38,15 +39,29 @@ export class OktaAuthGuard implements CanActivate {
       return true;
     }
 
+    /**
+     * Get the operation to perform on failed authentication from
+     * either the global config or route data injection.
+     */
+    const authGuardRedirect = route.data['authGuardRedirect'] || this.oktaAuth.getOnAuthRequired();
```
lboyette-okta: `authGuardRedirect` should probably be consistent in the places we allow extensibility (`onAuthRequired` or something)

packages/okta-angular/src/okta/okta.service.ts
```diff
@@ -84,12 +84,20 @@ export class OktaAuthService {
...
+    getOnAuthRequired(){
+      return this.config.getOnAuthRequired;
+    }
+
+    /**
      * Launches the login redirect.
      */
-    loginRedirect() {
+    loginRedirect(additionalParams?: object) {
       this.oktaAuth.token.getWithRedirect({
```
lboyette-okta: yeah, here we can check if the responseType has a string in it, then split it into an array.

packages/okta-angular/test/e2e/harness/src/app/app.module.ts
```diff
@@ -31,15 +48,20 @@ const appRoutes: Routes = [
     canActivate: [ OktaAuthGuard ]
   },
   {
-    path: 'login',
-    component: OktaLoginRedirectComponent
+    path: 'protected-with-data',
+    component: ProtectedComponent,
+    canActivate: [ OktaAuthGuard ],
+    data: {
+      authGuardRedirect: onFailedAuthGuard
```
lboyette-okta: idk if this will always be a redirect. We may consider renaming

packages/okta-angular/test/e2e/harness/src/app/sessionToken-login.component.ts
```diff
@@ -0,0 +1,49 @@
...
+      url: this.okta.getOktaAuth().options.url
+    });
+  }
+
+  signIn(username, password) {
+    this.oktaAuth.signIn({
+      username: username,
+      password: password
+    })
+    .then(res => {
```
lboyette-okta: for a fun ES6 shortening strategy (not necessary, but I think the airbnb linter throws an error), you can actually omit the extra brackets:

```
this.oktaAuth.signIn({
  username: username,
  password: password
})
.then(res => this.okta.loginRedirect({
  sessionToken: res.sessionToken
}))
.catch(err => console.log('Found an error', err));
```

packages/okta-angular/test/e2e/harness/src/app/sessionToken-login.component.ts
```diff
@@ -0,0 +1,49 @@
...
+      this.okta.loginRedirect({
+        sessionToken: res.sessionToken
+      });
+    })
+    .catch(err => {
+      console.log('Found an error', err);
+    })
+    .fail(err => {
```
lboyette-okta: `fail` isn't part of the Promise spec. Some Promise libraries added it to avoid problems in IE8/9, where `catch` is a reserved word. You've already defined `catch` here, so you can remove `fail`

packages/okta-angular/src/okta/okta.config.ts
robertjd: Yes, but necessary?  Someone opened issues because this was a limitation for them.

packages/okta-angular/test/e2e/harness/src/app/app.module.ts
jmelberg-okta: Updated to: `onAuthRequired`

packages/okta-angular/src/okta/okta.guard.ts
robertjd: Agreed that we should consistently use `onAuthRequired`

packages/okta-angular/src/okta/okta.service.ts
jmelberg-okta: Removing the ability to request a single token. Instead, we automatically return an `idToken` and `accessToken`

packages/okta-angular/src/okta/okta.service.ts
```diff
@@ -138,4 +151,19 @@ export class OktaAuthService {
       this.oktaAuth.tokenManager.clear();
       await this.oktaAuth.signOut();
     }
+
+    /**
+     * Scrub scopes to ensure 'openid' is included
+     */
+    scubScopes(scopes) {
```
robertjd: typo: scub

packages/okta-angular/README.md
```diff
@@ -83,9 +83,38 @@ Routes are protected by the `OktaAuthGuard`, which verifies there is a valid `ac
   component: MyProtectedComponent,
   canActivate: [ OktaAuthGuard ]
 }
-``` 
+```
+
+By default, the `OktaAuthGuard` will redirect the user to Okta for authentication. You can modify this functionality by adding a `data` attribute:
```
robertjd: We should move all of this custom-login-page content under a new heading "Using a Custom Login Page", this section should come before "Development".  The reasoning is (1) taking about the custom login page is a distraction from the "let's get the default case working" flow and (2) we need to elaborate on this custom login page case, as written this leaves them hanging.  We need to tell them that a custom login pages is intended to be used with the sign-in widget, and that they are responsible for using the widget to kick off the implicit redirect flow.  Keep this terse and link to SIW docs for more details.  After samples V2 strategy is complete we will also have a concrete example to link to :)

packages/okta-angular/README.md
```diff
@@ -83,9 +83,38 @@ Routes are protected by the `OktaAuthGuard`, which verifies there is a valid `ac
...
+    router.navigate(['/custom-login-page']);
+  }
+}
+```
+
+Alternativly, set this behavior globally by adding it to your configuration object:
+
+```typescript
+const onAuthRequired = ({oktaAuth, router}) => {
+  // Do something else
```
robertjd: Code comment here should be same as before: redirect to custom login page

packages/okta-angular/README.md
```diff
@@ -83,9 +83,38 @@ Routes are protected by the `OktaAuthGuard`, which verifies there is a valid `ac
...
+By default, the `OktaAuthGuard` will redirect the user to Okta for authentication. You can modify this functionality by adding a `data` attribute:
+
+```typescript
+path: 'protected',
+component: MyProtectedComponent,
+canActivate: [ OktaAuthGuard ],
+data: {
+  authGuardRedirect: ({oktaAuth, router}) => {
+    // Redirect the user to a custom login page
+    router.navigate(['/custom-login-page']);
```
robertjd: Why not use `/login` as the example?

packages/okta-angular/README.md
```diff
@@ -83,9 +83,38 @@ Routes are protected by the `OktaAuthGuard`, which verifies there is a valid `ac
...
+
+Alternativly, set this behavior globally by adding it to your configuration object:
+
+```typescript
+const onAuthRequired = ({oktaAuth, router}) => {
+  // Do something else
+  router.navigate(['/custom-login-page']);
+}
+
+const config = {
```
robertjd: In these code snippets, we should rename `config` to something less generic (so that when you see it in an isolated context like this, you know what you're dealing with).  This value is currently typed as `OktaConfig` so we should call it `oktaConfig`

packages/okta-angular/README.md
```diff
@@ -83,9 +83,38 @@ Routes are protected by the `OktaAuthGuard`, which verifies there is a valid `ac
...
+  issuer: environment.ISSUER,
+  ...
+  authGuardRedirect: onAuthRequired
```
robertjd: Should be `onAuthRequired: `

packages/okta-angular/README.md
jmelberg-okta: I wanted to hold off on adding explicit detail about a **custom login** page, since we will be updating our **[Okta + Angular + Sign-In Widget](https://okta.github.io/code/angular/okta_angular_sign-in_widget.html)** example to explain how this works.

Additionally, we have the problem of duplication in this README. It is nearly identical to the quickstart. As you mentioned, we should point the reader to our samples, quickstarts, and walkthroughs to avoid this.

PR #31, opened 2017-10-29
use babel-preset-env

### LINKED ISSUE #30: Unable to compile okta-react dependencies when using Webpack/Babel with Create-React-App
I'm looking for a way to use okta with my React app which was generated using Create-React-App, however, I'm running into trouble when adding these modules as they use async functions which are an ES2017 feature. By default, Webpack only transpiles my app code into ES5 using the babel loaders and plugins, however, it does not transpile external dependencies due to performance considerations. As a result, I am running into errors when it imports these modules. I believe this will be an issue for anyone using Create-React-App to bootstrap a new project, unless they eject and configure webpack to compile all dependencies as well (which is inadvisable as it drastically slows down webpack compilation times). I was able to correct the issue by transpiling the dependencies manually, but it seems like it would be much easier for consumers if these libraries were hosted in the more widely supported ES5 syntax. This seems like the standard for most other libraries. Do you know if this is a possibility or do you know of any workarounds for Create-React-App users?

jeroenransijn:
This should fix #30. I am not super comfortable with babel and Webpack. However, this does fix the async references in the dist folder.

### DIFF
--- a/packages/oidc-middleware/package.json
+++ b/packages/oidc-middleware/package.json
@@ -12,8 +12,16 @@
     "auth"
   ],
   "main": "index.js",
+  "config" : { 
+    "tck": {
+      "jar": "https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=com.okta.tests&a=okta-oidc-tck&v=0.2.0-SNAPSHOT&e=jar&c=shaded",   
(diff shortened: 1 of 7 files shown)


lboyette-okta:
Thanks for your contribution! It will be included in https://github.com/okta/okta-oidc-js/pull/51

Issue #28, opened 2017-10-25
Config scopes ignored when calling OktaAuthService.loginRedirect()


RaduGrama:
`OktaAuthService.loginRedirect()` is designed to pass along, by default, the `OktaConfig.scopes` value passed in to `OktaAuthModule.initAuth()`, but that is not functioning as intended and as a consequence the scopes default to `['openid', 'email']`.

The code below refers to this.config.scope, which is always 'undefined': 

https://github.com/okta/okta-oidc-js/blob/d7f4d9b769dee83058a6e7489673b5bf17e78354/packages/okta-angular/src/okta/okta.service.ts#L104

`OktaAuthService` constructor initializes this.config with auth, and auth.scopes is initialized right above via a call to `this.scrubScopes()`:

https://github.com/okta/okta-oidc-js/blob/d7f4d9b769dee83058a6e7489673b5bf17e78354/packages/okta-angular/src/okta/okta.service.ts#L52-L60

One issue is that the string returned by `OktaAuthService.scrubScopes()` is assigned to `this.config.scopes` instead of `this.config.scope`.

The second issue is that even if the assignment in line 55 is changed to: `auth.scopes = this.scrubScopes(auth.scopes);`, the subsequent call to the underlying JS library will fail, as `scope` is obsolete and the current code simply assigns `scope` (a string) to `scopes` (an array of strings), and as a consequence the code further fails because `join()` does not exist on a string.

The solution would be to:

- Change `OktaAuthService.scrubScopes()` to return a scrubbed array of strings instead of a string;
- Change line 104 to: `scopes: this.config.scopes`

The workarounds I found so far:

- When calling `OktaAuthService.loginRedirect()`, include `scopes: [<desired scopes>]` as part of the `additionalParams` parameter
- Or, before the line containing the call to `OktaAuthService.loginRedirect()`, call `oktaAuthService.getOktaConfig().scope = [<desired scopes>]` - this just takes advantage of the bug, so it should be least desired workaround

PR #32, opened 2017-10-30
:bug: Fixes overwriting scopes

jmelberg-okta:
- :bug: Fixes a bug where injected scopes would be overwritten due to referencing an `undefined` value.

Resolves: OKTA-146916
Resolves: #28


jakehockey10:
When will this be released?  :(

jrowe88:
Ran into this bug yesterday in fact!  Without being able to specify scopes this package is kind of useless to us.  Can you publish an npm update soon?

PR #45, opened 2017-11-13
Cast passport strategy errors and 401 responses

robertjd:
This changes the default error code from 500 to 401, which technically is a breaking change.

I'm adding the Alpha label to this library until it becomes 1.x


packages/oidc-middleware/src/connectUtil.js
```diff
@@ -35,6 +35,11 @@ connectUtil.createOIDCRouter = context => {
   } = context.options.routes;
   oidcRouter.use(loginPath, bodyParser.urlencoded({ extended: false}), connectUtil.createLoginHandler(context));
   oidcRouter.use(callbackPath, connectUtil.createCallbackHandler(context));
+  oidcRouter.use(function(err, res, res, next){
```
lboyette-okta: just for consistency, we should be using the `() => {}` syntax for anonymous functions. Sounds like something I should add to our linter

PR #51, opened 2017-11-22
Use babel-env and add babel transform runtime

lboyette-okta:


packages/okta-react/.babelrc
```diff
@@ -1,11 +1,17 @@
 {
-  "presets": ["es2015", "react"],
-  "plugins": ["transform-object-rest-spread", "transform-class-properties"],
-  "env": {
-    "development": {
-      "plugins": [
-        "babel-plugin-dynamic-import-node"
+  "presets": [["env", {
+    "targets": {
+      "browsers": [
```
jmelberg-okta: Curious why we specify the configuration here, as opposed to import from a shared `browserslist` config file. I'd assume devs have production and development browser targets already, so it might make sense to use their config instead of hardcode our own.

Reference: [BrowsersList Config File](https://github.com/ai/browserslist#config-file)

lboyette-okta: This will be compiled in our environment, before the developer pulls in the library. That means we'll never see their config.

I use the same browser support we have for the widget: https://github.com/okta/okta-signin-widget/blob/master/package.json#L17-L19

jeroenransijn:
When is this going live? I am still seeing `async` when I install `0.0.7`.

PR #73, opened 2017-12-12
Updating package versions

robertjd:


packages/oidc-middleware/notes.md
```diff
@@ -0,0 +1 @@
+what should the default session options be?
```
bretterer: Maybe you want to remove the notes.md file?

robertjd: oops, thanks

PR #76, opened 2017-12-14
:seedling: Adds Vue.js plugin for Okta Auth

jmelberg-okta:
### Description
Adds plugin for Vue applications to easily use `okta-auth-js`

For running/setup instructions, see the [README](https://github.com/okta/okta-oidc-js/blob/jm-vue/packages/okta-vue/README.md).


packages/okta-vue/README.md
```diff
@@ -0,0 +1,142 @@
+# Okta Vue SDK
+The Okta Vue SDK is a wrapper around [Okta Auth JS](https://github.com/okta/okta-auth-js), that builds on top of Okta's [OpenID Connect API](https://developer.okta.com/docs/api/resources/oidc.html).
+
+This library currently supports:
+  - [OAuth 2.0 Implicit Flow](https://tools.ietf.org/html/rfc6749#section-1.3.2)
+
+## Getting Started
+* If you do not already have a **Developer Edition Account**, you can create one at [https://developer.okta.com/signup/](https://developer.okta.com/signup/).
+* If you don't have a Vue app, or are new to Vue, please start with the [Vue Quickstart](https://vuejs.org/v2/guide/#Getting-Started) guide. It will walk you through the creation of a Vue app, creating routes, and other application development essentials.
```
robertjd: This Vue quickstart is a naive html+js app, aka not a webpack app that you can easily add node modules to.  Is there a better place to point folks, so that they're read to install our lib as a node module?

robertjd:
I haven't dug into the code yet, but I have put together a quick sample app and used the library by following the README.  A few things so far:

* How can I wire up a Login button to kick off the login flow? In our other libs we attach something to `/login` that will kick off the redirect if you navigate there.
* How do i get the access token and use it?

bretterer:
When having this included in your javascript without config, it throws errors in the console about `cannot read property 'issuer' of undefined`.  We should make sure to check the item first.  This was the case of not having a config set, so maybe we can update this to see if there is no config, we provide a console warning of something like `console.warning('I see you have added okta-js but did not configure it yet.  For information on how to configure it, go to .... blah blah blah..')`

packages/okta-vue/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+```typescript
+// router/index.js
+
+import Auth from '@okta/okta-vue'
+
+Vue.use(Auth, {
+  issuer: 'https://{yourOktaDomain}.com/oauth2/default',
+  client_id: '{client_id}',
+  redirect_uri: 'https://localhost:{port}/implicit/callback',
```
bretterer: For the redirect_uri, we are using `https` in this guide, but when you go through the setup in the okta dashboard, it defaults to `http`.  We should set it up to be the same, either way

jmelberg-okta: Good catch!

packages/okta-vue/README.md
```diff
@@ -0,0 +1,223 @@
...
+    <router-view/>
+  </div>
+</template>
+
+<script>
+
+export default {
+  name: 'app',
+  data: function () {
+    return { authenticated: false }
```
robertjd: Styling: let's add a newline for the properties here

packages/okta-vue/README.md
```diff
@@ -0,0 +1,223 @@
...
+  },
+  created () { this.isAuthenticated() },
```
robertjd: Styling: let's add a newline for this statment

packages/okta-vue/README.md
```diff
@@ -0,0 +1,223 @@
...
+  watch: {
+    // Everytime the route changes, check for auth status
+    '$route': 'isAuthenticated'
```
robertjd: Watching '$route' wasn't causing any hooks for me?  I ended up adding a call to `this.isAuthenticated()` below `await this.$auth.logout()`

packages/okta-vue/README.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+### Reference
+**Configuration Options**
+  - `issuer` **(required)**: The OpenID Connect `issuer`
+  - `client_id` **(required)**: The OpenID Connect `client_id`
+  - `redirect_uri` **(required)**: Where the callback is hosted
+  - `scopes` *(optional)*: Reserved or custom claims to be returned in the tokens
+
+### Using a custom login-page
```
robertjd: Let's add a section that enumerates the available API methods, like what we have here: https://github.com/okta/okta-oidc-js/tree/master/packages/okta-react#auth

packages/okta-vue/src/utils/Utils.js
```diff
@@ -0,0 +1,22 @@
...
+  localStorage.removeItem('referrerPath')
+  return path
+}
+
+const Utils = {
+  initConfig,
+  getFromUri
+}
+
+export { Utils }
```
robertjd: Is this utils file necessary?  These methods are only called from a single place each, so I'm not sure it's worth breaking them out like this?

packages/okta-vue/src/components/ImplicitCallback.vue
```diff
@@ -0,0 +1,10 @@
+<template></template>
```
robertjd: Is it necessary to define an entire component for this callback handler?  It appears you can use an empty render() function to bypass this need.

packages/okta-vue/src/utils/Utils.js
```diff
@@ -0,0 +1,22 @@
...
+  if (!auth.issuer) missing.push('issuer')
+  if (!auth.client_id) missing.push('client_id')
+  if (!auth.redirect_uri) missing.push('redirect_uri')
+  if (!auth.scopes) auth.scopes = ['openid']
+  if (missing.length) throw new Error(`${missing.join(', ')} must be defined`)
+  return auth
+}
+
+const getFromUri = () => {
+  const path = localStorage.getItem('referrerPath') || '/'
```
robertjd: "referrer" feels a bit overloaded.  I feel like we should be more descriptive, something like `postLoginPath`

packages/okta-vue/src/components/ImplicitCallback.vue
```diff
@@ -0,0 +1,10 @@
...
+
+<script>
+export default {
+  name: 'ImplicitCallback',
+  async beforeMount () {
+    this.$router.push({ path: await this.$auth.handleAuthentication() })
```
robertjd: The method `$auth.handleAuthentication()` is returning a URL string, which is not obvious from the name of this method.  I think it would be better if this method did not return any value, and instead handle the URL need separately.  Something like:

```
await this.$auth.handleAuthentication()
this.$auth.postLoginRedirect()
```

packages/okta-vue/src/components/ImplicitCallback.vue
jmelberg-okta: Good find. I've updated the export function. 👍

packages/okta-vue/README.md
jmelberg-okta: We need both in this case. On route change, we need to call `this.$auth.isAuthenticated()` to ensure the DOM is updated correctly. Looks like `logout` had some issues and are fixed in the next commit.

packages/okta-vue/src/utils/Utils.js
jmelberg-okta: This is a design decision we made earlier to be used across all of our SDKs. Noted [in this comment](https://github.com/okta/okta-oidc-js/pull/11#discussion-diff-141099927R26).

We can create a follow-up issue to change this if you'd like, but this value shouldn't be used/modified by the end developer.

packages/okta-vue/package.json
```diff
@@ -0,0 +1,68 @@
+{
+  "name": "@okta/okta-vue",
+  "version": "0.1.0",
```
lboyette-okta: is this the version we want?

packages/okta-vue/src/OktaAuth.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function handleCallback () { return ImplicitCallback }
+
+const initConfig = auth => {
+  const missing = []
+  if (!auth.issuer) missing.push('issuer')
+  if (!auth.client_id) missing.push('client_id')
+  if (!auth.redirect_uri) missing.push('redirect_uri')
+  if (!auth.scopes) auth.scopes = ['openid']
```
lboyette-okta: It's small, but I like the idea of sticking to the oauth spec with `auth.scope` as a string. It makes it easier to move to a different lib in the future.

packages/okta-vue/src/OktaAuth.js
```diff
@@ -0,0 +1,77 @@
...
+  if (missing.length) throw new Error(`${missing.join(', ')} must be defined`)
```
lboyette-okta: This is awesome! Fail once, and know all the reasons you failed

packages/okta-vue/src/OktaAuth.js
jmelberg-okta: Done. Good catch.

packages/okta-vue/package.json
jmelberg-okta: I figured this is a good starting point. Do you or @robertjd have a preference? I'll defer to you both.

packages/okta-vue/src/OktaAuth.js
robertjd: Ditto, let's make string the norm across the board

packages/okta-vue/package.json
lboyette-okta: That's fine with me. Our other libs use the patch version, but everything is independent now so I think we should do what makes sense for this lib.

packages/okta-vue/README.md
```diff
@@ -52,7 +52,7 @@ Vue.use(Auth, {
   issuer: 'https://{yourOktaDomain}.com/oauth2/default',
   client_id: '{client_id}',
   redirect_uri: 'http://localhost:{port}/implicit/callback',
-  scopes: ['openid', 'profile', 'email']
+  scope: ['openid', 'profile', 'email']
```
lboyette-okta: this should show scope as a string

packages/okta-vue/src/OktaAuth.js
```diff
@@ -69,7 +69,7 @@ const initConfig = auth => {
   if (!auth.issuer) missing.push('issuer')
   if (!auth.client_id) missing.push('client_id')
   if (!auth.redirect_uri) missing.push('redirect_uri')
-  if (!auth.scopes) auth.scopes = ['openid']
+  if (!auth.scope) auth.scope = ['openid']
```
lboyette-okta: this should be `if (!auth.scope) auth.scope = 'openid'`

packages/okta-vue/test/e2e/harness/src/router/index.js
```diff
@@ -10,7 +10,7 @@ Vue.use(Auth, {
   issuer: process.env.ISSUER,
   redirect_uri: process.env.REDIRECT_URI,
   client_id: process.env.CLIENT_ID,
-  scopes: ['openid', 'profile', 'email']
+  scope: ['openid', 'profile', 'email']
```
lboyette-okta: `scope: 'openid profile email'`

packages/okta-vue/src/OktaAuth.js
```diff
@@ -0,0 +1,77 @@
+import * as AuthJS from '@okta/okta-auth-js'
```
lboyette-okta: in the other libs, we call this file `Auth.js` and the lib `OktaAuth`

packages/okta-vue/src/OktaAuth.js
```diff
@@ -0,0 +1,77 @@
...
+    issuer: authConfig.issuer,
+    redirectUri: authConfig.redirect_uri,
+    url: authConfig.issuer.split('/oauth2/')[0]
+  })
+
+  Vue.prototype.$auth = {
+    loginRedirect (additionalParams) {
+      return oktaAuth.token.getWithRedirect({
+        responseType: ['id_token', 'token'],
+        scopes: authConfig.scope,
```
lboyette-okta: here, we can split `scope` by spaces

PR #81, opened 2018-01-06
Added string[] as valid type for scope in config

matthewstrom:
The actual JavaScript for `OktaAuthModule` accepts a string array for the `scope` property of the config; however, the faulty type definition causes compilation to fail when `scope` is an array.

### DIFF
--- a/packages/okta-angular/src/okta/okta.config.ts
+++ b/packages/okta-angular/src/okta/okta.config.ts
@@ -16,7 +16,7 @@ export interface OktaConfig {
   issuer?: string;
   redirectUri?: string;
   clientId?: string;
-  scope?: string;
+  scope?: string | string[];
   onAuthRequired?: Function;
 }
 


robertjd:
Thanks for this PR!  We’re actually trying to move to a world where scope is always defined as a string (per OAuth norms) so we want to leave this as a string. Ideally, we should only be checking for the only `String` type [here](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-angular/src/okta/okta.service.ts#L162:L165). If you’d like to update this PR to remove that snippet instead, we will get this merged in ASAP!

matthewstrom:
Is that to say that a space-delimited list of scopes will work?

By "remove that snippet", are you proposing that line 162–165 in _okta.service.ts_ should be removed? Won't that break backward compatibility?

Perhaps some better documentation is in order? Something that indicates that `scope` accepts a space-delimited list.

matthewstrom:
Is that to say that a space-delimited list of scopes will work?

By "remove that snippet", are you proposing that line 162–165 in _okta.service.ts_ should be removed? Won't that break backward compatibility?

Perhaps some better documentation is in order? Something that indicates that `scope` accepts a space-delimited list.

cmckni3:
I modified the parameter of `scrubScopes` in #116. Do you want me to remove the `string[]` type and the related code @robertjd?

robertjd:
Thanks all for pointing this out, in 1.0 we're going to make this string-only.  See #120

jmelberg-okta:

PR #83, opened 2018-01-08
Update publish script

jmelberg-okta:
Currently`execSync` script throws an error if a package doesn't exist in the npm registry. Surrounding this with a simple `try catch` will catch the `404` error and publish it as a new version.

### DIFF
--- a/scripts/publish_registry.js
+++ b/scripts/publish_registry.js
@@ -26,7 +26,13 @@ dirs.forEach(name => {
   const moduleWithVersion = `${pkg.name}@${pkg.version}`;
 
   console.log(`Checking if ${moduleWithVersion} exists`);
-  const isInPublicNpm = !!execSync(`npm view ${moduleWithVersion} --registry ${registry}`).toString();
+
+  try {
+    const isInPublicNpm = !!execSync(`npm view ${moduleWithVersion} --registry ${registry}`).toString();
+  } catch (err) {
+    // We expect packages that do not exist to throw a 404 error
+  }
+
   if (isInPublicNpm) {
     console.log(`${moduleWithVersion} exists`);
   } else {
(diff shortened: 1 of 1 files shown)


scripts/publish_registry.js
```diff
@@ -26,7 +26,13 @@ dirs.forEach(name => {
   const moduleWithVersion = `${pkg.name}@${pkg.version}`;
 
   console.log(`Checking if ${moduleWithVersion} exists`);
-  const isInPublicNpm = !!execSync(`npm view ${moduleWithVersion} --registry ${registry}`).toString();
+
+  try {
+    const isInPublicNpm = !!execSync(`npm view ${moduleWithVersion} --registry ${registry}`).toString();
```
lboyette-okta: `const` makes a variable block-specific, so it won't exist by line 36. Use `let` to define `isInPublicNpm` outside of the `try` block, then assign it inside the block

scripts/publish_registry.js
```diff
@@ -26,7 +26,13 @@ dirs.forEach(name => {
...
+  } catch (err) {
+    // We expect packages that do not exist to throw a 404 error
```
lboyette-okta: we should use console.log what we think happened

```console.log(`${pkg.name} does not have any published versions`);```

File: scripts/publish_registry.js [hunk 3]
```diff
@@ -27,10 +27,12 @@ dirs.forEach(name => {
 
   console.log(`Checking if ${moduleWithVersion} exists`);
 
+  let isInPublicNpm = false;
   try {
-    const isInPublicNpm = !!execSync(`npm view ${moduleWithVersion} --registry ${registry}`).toString();
+    isInPublicNpm = !!execSync(`npm view ${moduleWithVersion} --registry ${registry}`).toString();
   } catch (err) {
     // We expect packages that do not exist to throw a 404 error
+    console.log(`${pkg.name} does not have have any published versions`);
```
lboyette-okta: `have have` => `have`

jmelberg-okta: Doh.

PR #95, opened 2018-01-17
Add tokenManager to Auth.js [React]

adamjcsmith:
This PR extends the config object in Auth.js to accept `tokenManager`, falling back to the default object properties (as defined in [TokenManager.js](https://github.com/okta/okta-auth-js/blob/master/lib/TokenManager.js#L134)) if not specified. Supported in the [OktaAuth spec](https://github.com/okta/okta-auth-js/blob/master/lib/clientBuilder.js#L141).

### DIFF
--- a/packages/okta-react/README.md
+++ b/packages/okta-react/README.md
@@ -178,6 +178,12 @@ Security is the top-most component of okta-react. This is where most of the conf
     1. `auth.login` is called
     2. SecureRoute is accessed without authentication
 
+* **tokenManager** (optional)
+  Specify where the Okta token will be stored. Default is `localStorage`.
+  * `localStorage`
+  * `sessionStorage`
+  * `cookie`
+
 #### Example
 
 ```typescript
--- a/packages/okta-react/README.md
+++ b/packages/okta-react/README.md
@@ -179,10 +179,13 @@ Security is the top-most component of okta-react. This is where most of the conf
     2. SecureRoute is accessed without authentication
 
 * **tokenManager** (optional)
-  Specify where the Okta token will be stored. Default is `localStorage`.
-  * `localStorage`
-  * `sessionStorage`
-  * `cookie`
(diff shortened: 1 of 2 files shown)


robertjd:
Hi @adamjcsmith , thanks for the PR!  We'll take a look at this.  Which option is it that you're needing to set, and tell me a bit about the case you're trying to solve?  I ask because we've been trying to hide the details of the Auth SDK behind this library, so rather than having the options pass through directly we may want to have an interface on top of it.  Thanks for the input!

adamjcsmith:
Hey @robertjd , thanks for reviewing. The option I needed to set was to store the access/id tokens in the form of cookies, rather than the default (localStorage).

Understand the abstraction, perhaps you could provide a new prop, like `tokenStorage`, accepting one of the allowed strings.

robertjd:
Hi @blainekasten , thanks for the comments.  We are going to add the ability to configure the token manager from this library 👍 

Regarding cookies vs. local storage, that conversation boils down to the XSS attack vector.  Yes, local storage is readable by JavaScript.  But so are cookies, unless you specify `httpOnly` on a cookie that is set by an HTTP response.  In this library we are working with the OIDC implicit flow, so we get our tokens back in the URL and then the client is responsible for storing them, which means we can't take advantage of `httpOnly`.  As such, local storage or cookies have the same vulnerabilities in this situation.

When dealing with tokens stored client-side (regardless of the storage method), the important thing is to make sure your app is not vulnerable to XSS. That means sanitizing all text input and using Content-Security-Policy. OWASP has great info here: https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet

manueltanzi-okta:
@adamjcsmith We just released:
- okta-angular v1.1.0
- okta-react v1.2.0
- okta-vue v1.1.0
Which add the possibility to configure TokenManager parameters, check it out!

Issue #19, opened 2017-10-01
okta-react: Make Auth.js service available outside of React components


alaister:
Hi,
It would be useful if you made the methods in the Auth class available without having to use the withAuth HOC.

This is useful when you need to use the access token outside of React. For example when using [Apollo Client](http://dev.apollodata.com/react/auth.html#Header).

Thanks

leandroleo02:
Hi!

I have a similar issue, any workaround?

Thanks!

gaastonsr:
Same here but with Vue.

robertjd:
This should now be available in 0.0.11, as resolved by #97.  Please give it a try and let us know if it works.

PR #97, opened 2018-01-17
Add Auth access to react

lboyette-okta:
related-to: #19


packages/okta-react/README.md
```diff
@@ -173,6 +173,7 @@ Security is the top-most component of okta-react. This is where most of the conf
 * **client_id** (required) - The OpenId Connect `client_id`
 * **redirect_uri** (required) - Where the callback handler is hosted
 * **onAuthRequired** (optional)
+* **auth** (optional) - Provide an Auth object instead of the options above
```
robertjd: Let's add a note about why this is useful, i.e. when you need to share the auth instance w/ other code, e.g. code that needs to get the access token

jasonrberk: it would be _really_ nice to see a simple example....like in your API.js service to get your token.  I just spent two days figuring it out

PR #108, opened 2018-01-31
Enabling travis CI

vijetmahabaleshwar-okta:


.travis.yml
```diff
@@ -1 +1,18 @@
...
+language: node_js
+
+node_js:
+- '8.4.0'
+
+addons:
+  chrome: stable
+  apt:
+    packages:
+      - oracle-java8-set-default
```
vijetmahabaleshwar-okta: Added java package as nightwatch tests used in okta-vue library needs java, apparently.
The tests are still failing on travis. Investigating...

packages/jwt-verifier/package.json
```diff
@@ -7,14 +7,16 @@
   "main": "lib.js",
   "config": {
     "tck": {
-      "jar": "https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=com.okta.oidc.tck&a=okta-oidc-tck&v=0.2.1-SNAPSHOT&e=jar&c=shaded",
-      "setup_env": "cross-env ISSUER=http://localhost:9090/oauth2/default cross-env CLIENT_ID=OOICU812 cross-env CLIENT_SECRET=VERY_SECRET",
-      "command": "java -Dconfig=./test/integration-test/resources/testRunner.yml -cp ./okta-oidc-tck-0.2.1-SNAPSHOT-shaded.jar org.testng.TestNG -d target/cli-test-output ./test/integration-test/resources/testng.xml"
+      "jar": "https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=com.okta.oidc.tck&a=okta-oidc-tck&v=0.2.2-SNAPSHOT&e=jar&c=shaded",
+      "setup_env": "export ISSUER=http://localhost:9090/oauth2/default export CLIENT_ID=OOICU812 export CLIENT_SECRET=VERY_SECRET",
```
vijetmahabaleshwar-okta: Can't use `cross-env` as we start another process when TCK is run, and `cross-env` doesn't set env vars across processes.

packages/jwt-verifier/test/integration-test/resource-server.js
```diff
@@ -79,11 +79,20 @@ app.get('/secure', authenticationRequired, (req, res) => {
  * print some messages for the user if they are authenticated
  */
 app.get('/api/messages', authenticationRequired, (req, res) => {
-  res.json([{
-    message: 'Hello, word!'
-  }]);
+  res.json({
```
vijetmahabaleshwar-okta: Updated as our TCK tests expect `messages` instead of a single `message`

packages/jwt-verifier/test/integration-test/resources/testRunner.yml
```diff
@@ -17,21 +17,10 @@ scenarios:
   implicit-flow-local-validation:
     disabledTests:
       - wrongScopeAccessTokenTest
+      - scopeAccessTokenTest
```
vijetmahabaleshwar-okta: This test will be enabled after a new TCK version is released. It was failing due to an error in groovy test. Test updated but TCK jar needs to bumped up.

packages/oidc-middleware/test/e2e/conf.js
```diff
@@ -20,7 +20,8 @@ exports.config = {
       args: [
         '--headless',
         '--disable-gpu',
-        '--window-size=800,600'
+        '--window-size=800,600',
+        '--no-sandbox'
```
vijetmahabaleshwar-okta: Without `--no-sandbox` arg, chrome driver crashes on travis.

scripts/travis.sh
```diff
@@ -0,0 +1,64 @@
...
+
+if (cat ./okta-react-test.log | grep -e "FAILED:" -e "Failed:" > /dev/null); then
+    echo "[ERROR] okta-react tests failed!"
+    cat ./okta-react-test.log
+    exit -1
+else
+    echo "okta-react tests successful."
+fi
+
+# export DBUS_SESSION_BUS_ADDRESS=/dev/null
```
vijetmahabaleshwar-okta: okta-vue tests are failing as selenium webdriver process isn't started. The issue doesn't occur locally.
Read that it can be a version mismatch problem or a problem with running in docker. Will enable once the error is fixed.

packages/oidc-middleware/package.json
```diff
@@ -14,14 +14,17 @@
   "main": "index.js",
   "config": {
     "tck": {
-      "jar": "https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=com.okta.oidc.tck&a=okta-oidc-tck&v=0.2.1-SNAPSHOT&e=jar&c=shaded",
-      "setup_env": "cross-env ISSUER=http://localhost:9090/oauth2/default cross-env CLIENT_ID=OOICU812 cross-env CLIENT_SECRET=VERY_SECRET",
-      "command": "java -cp ./test/integration-test/resources/:./okta-oidc-tck-0.2.1-SNAPSHOT-shaded.jar org.testng.TestNG -d target/cli-test-output ./test/integration-test/resources/testng.xml"
+      "jar": "https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=com.okta.oidc.tck&a=okta-oidc-tck&v=0.2.2-SNAPSHOT&e=jar&c=shaded",
+      "setup_env": "export ISSUER=http://localhost:9090/oauth2/default export CLIENT_ID=OOICU812 export CLIENT_SECRET=VERY_SECRET",
```
vijetmahabaleshwar-okta: `cross-env` sets env vars only in the current process. But TCK test starts the server as another process which doesn't have access to their env vars.

scripts/travis.sh
```diff
@@ -0,0 +1,65 @@
...
+
+if (cat ./jwt-verifier-test.log | grep -i -e "failed:" -e "failing" > /dev/null); then
+    echo "[ERROR] jwt-verifier tests failed!"
+    cat ./jwt-verifier-test.log
+    exit -1
+else
+    echo "jwt-verifier tests successful."
+fi
+
+export CLIENT_ID=0oad5983foWn2Gefr0h7
```
jmelberg-okta: Since these are shared, we should put them into a travis environment var.

vijetmahabaleshwar-okta: Actually, since we are testing both auth code and implicit flows, and even within these flows, some apps use port 8080, while some others 3000, we can't have just 1 client Id. Will need to set this env var before each test. 
I already have a CLIENT_ID travis env var, but that is for a web app.

packages/okta-react/package.json
```diff
@@ -12,7 +12,8 @@
     "prestart": "npm run build && npm run build:harness",
     "start": "npm run --prefix test/e2e/harness start",
     "pretest": "npm run build && npm run build:harness",
-    "test": "npm run --prefix test/e2e/harness test"
+    "test": "npm run --prefix test/e2e/harness test",
+    "install:package": "npm install"
```
robertjd: It this `install:package` needed?

scripts/travis.sh
```diff
@@ -0,0 +1,65 @@
...
+
+npm run install:packages
+
+export CLIENT_ID=0oad5a19u0IVnKU1g0h7
+export REDIRECT_URI=http://localhost:8080/implicit/callback
+
+cd packages/jwt-verifier
+npm test > ./jwt-verifier-test.log
+
+if (cat ./jwt-verifier-test.log | grep -i -e "failed:" -e "failing" > /dev/null); then
```
robertjd: Why are we manually doing this error handling logic?  Shouldn't the `npm test` command bubble it's error upwards?  Also, why are we manually invoking tests rather than using `lerna run test` and pushing the test configuration down to each package?

packages/oidc-middleware/package.json
```diff
@@ -14,14 +14,17 @@
...
+      "command": "java -cp ./test/integration-test/resources/:./okta-oidc-tck-0.2.2-SNAPSHOT-shaded.jar org.testng.TestNG -d target/cli-test-output ./test/integration-test/resources/testng.xml"
     }
   },
   "scripts": {
-    "test": "mocha test/unit/*.js && webdriver-manager update && protractor test/e2e/conf.js",
-    "tck-test": "(ls ./okta-oidc-tck-0.2.1-SNAPSHOT-shaded.jar || curl $npm_package_config_tck_jar -L -o okta-oidc-tck-0.2.1-SNAPSHOT-shaded.jar) && $npm_package_config_tck_setup_env && mkdir -p target && $npm_package_config_tck_command || echo done"
+    "test": "npm run test:unit && npm run test:e2e && npm run test:integration",
+    "test:e2e": "webdriver-manager update --gecko false && protractor test/e2e/conf.js",
+    "test:integration": "(ls ./okta-oidc-tck-0.2.2-SNAPSHOT-shaded.jar || curl $npm_package_config_tck_jar -L -o okta-oidc-tck-0.2.2-SNAPSHOT-shaded.jar) && $npm_package_config_tck_setup_env && mkdir -p target && $npm_package_config_tck_command || echo done",
```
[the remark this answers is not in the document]
robertjd: I'd prefer to not bring another tool in unless it's really necessary.

packages/jwt-verifier/package.json
```diff
@@ -7,14 +7,15 @@
   "main": "lib.js",
   "config": {
     "tck": {
-      "jar": "https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=com.okta.oidc.tck&a=okta-oidc-tck&v=0.2.1-SNAPSHOT&e=jar&c=shaded",
-      "setup_env": "cross-env ISSUER=http://localhost:9090/oauth2/default cross-env CLIENT_ID=OOICU812 cross-env CLIENT_SECRET=VERY_SECRET",
-      "command": "java -Dconfig=./test/integration-test/resources/testRunner.yml -cp ./okta-oidc-tck-0.2.1-SNAPSHOT-shaded.jar org.testng.TestNG -d target/cli-test-output ./test/integration-test/resources/testng.xml"
+      "jar": "https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=com.okta.oidc.tck&a=okta-oidc-tck&v=0.2.2-SNAPSHOT&e=jar&c=shaded",
+      "command": "java -Dconfig=./test/integration-test/resources/testRunner.yml -cp ./okta-oidc-tck-0.2.2-SNAPSHOT-shaded.jar org.testng.TestNG -d target/cli-test-output ./test/integration-test/resources/testng.xml"
```
jmelberg-okta: We should try to make `okta-oidc-tck-*` a variable, that way we only need to replace this once when we update the version

packages/oidc-middleware/package.json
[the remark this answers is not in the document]
jmelberg-okta: I think we should consider it (or a similar framework). This would help us differentiate tests between our JS frameworks (Angular, React, Vue) and other libraries. We can chat about this a bit more later.

packages/okta-react/package.json
jmelberg-okta: 😕

packages/okta-vue/test/e2e/harness/package.json
```diff
@@ -6,14 +6,16 @@
   "private": true,
   "scripts": {
     "dev": "webpack-dev-server --inline --progress --config build/webpack.dev.conf.js",
-    "start": "npm run dev",
+    "start": "npm run env && npm run dev",
+    "env": "node ./scripts/prebuild.js",
```
jmelberg-okta: nit: To avoid the `&&` operator, we should make this a `prestart` and `pretest` script.

scripts/travis.sh
jmelberg-okta: Don't we have control over these apps? We should be able to use the default (or fallback) callback URIs instead of hardcoding these. Additionally, I'd prefer to use environment vars if possible.

scripts/travis.sh
jmelberg-okta: +1.

We could check for any non-zero exit code by using `||` :
`npm test || echo "Tests Failed!"`

scripts/travis.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Disregard this for review
+# export CLIENT_ID=0oadwuob65038lE6f0h7
+# export REDIRECT_URI=http://localhost:3000/implicit/callback
+
+# cd packages/okta-vue/test/e2e/harness
+# npm install
+# npm install chromedriver@2.33.0 --chromedriver-cdnurl=http://chromedriver.storage.googleapis.com
+# curl https://selenium-release.storage.googleapis.com/3.5/selenium-server-standalone-3.5.0.jar > selenium-server-standalone.jar
+# npm test
```
jmelberg-okta: Why is this failing in travis but passing locally? Can you attach logs to this for us to look at please?

packages/oidc-middleware/package.json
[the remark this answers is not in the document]
robertjd: Ok, we can keep chatting but I don't think we should hold up this PR.

packages/okta-vue/test/e2e/harness/package.json
robertjd: IMO I actually find it harder to fish through the scripts map and find all the pre/post scripts.  I know we do this everywhere and that it's a node thing so I'm not saying we should change course, but just want to put my .02 in

packages/jwt-verifier/package.json
vijetmahabaleshwar-okta: Moving this to a separate script which would solve this problem

packages/oidc-middleware/package.json
vijetmahabaleshwar-okta: Moving this to a script

packages/okta-react/package.json
vijetmahabaleshwar-okta: This was related to some tests having local node_modules dependencies.
Since we `hoist = true` in `lerna.json`, these dependencies would be installed in the root `node_modules` which would cause these tests to fail.
As discussed with @robertjd, I'm removing `hoist = true` after which we can remove `install:package`

packages/okta-vue/test/e2e/harness/package.json
jmelberg-okta: Yeah - I think I was a bit confused by what `env` was doing by the name. Maybe we can rename it to: `setup-env-vars`?

scripts/travis.sh
vijetmahabaleshwar-okta: I read that it's because of some issue with running nightwatch tests on docker container. Will attach the logs.

vijetmahabaleshwar-okta: @jmelberg-okta: This is the error I get on travis - 

```
Starting selenium server... started - PID:  6133
[Test] Test Suite
=====================
Running:  redirects to Okta for login when trying to access a protected page
Error retrieving a new session from the selenium server
Connection refused! Is selenium server started?
{ Error: socket hang up
    at createHangUpError (_http_client.js:345:15)
    at Socket.socketCloseListener (_http_client.js:377:23)
    at emitOne (events.js:120:20)
    at Socket.emit (events.js:210:7)
    at TCP._handle.close [as _onclose] (net.js:545:12) code: 'ECONNRESET' }
```

packages/okta-angular/test/e2e/harness/scripts/prebuild.js
```diff
@@ -30,8 +28,7 @@ const defaultEnvValues = {
   BASE_URI: 'http://localhost:3000',
   REDIRECT_URI: 'http://localhost:3000/implicit/callback',
   ISSUER: 'https://{yourOktaDomain}.com/oauth2/default',
-  CLIENT_ID: '{clientId}',
-  CLIENT_SECRET: '{clientSecret}',
```
jmelberg-okta: I think we're going to need to remove this from `src/environments/environments.ts.environments` so our `prebuild.js` script doesn't error looking for it in the template.

scripts/travis.sh
jmelberg-okta: Gotcha. Lets investigate this as a follow up.

packages/okta-vue/test/e2e/harness/package.json
```diff
@@ -6,14 +6,16 @@
...
     "e2e": "node test/e2e/runner.js",
-    "test": "npm run e2e",
+    "test": "npm run env && npm run e2e",
     "lint": "eslint --ext .js,.vue src test/e2e/specs",
     "build": "node build/build.js"
   },
   "dependencies": {
     "@okta/okta-auth-js": "^1.11.0",
+    "chromedriver": "^2.33.0",
```
jmelberg-okta: Looks like this got changed back. Any reason why?

packages/okta-vue/test/e2e/harness/scripts/prebuild.js
```diff
@@ -0,0 +1,46 @@
...
+const targetEnvironmentTemplateFileName = 'dev.env.js.template';
+const targetEnvironmentFileName = 'dev.env.js';
+
+const defaultEnvValues = {
+  PORT: 3000,
+  BASE_URI: 'http://localhost:3000',
+  REDIRECT_URI: 'http://localhost:3000/implicit/callback',
+  ISSUER: 'https://{yourOktaDomain}.com/oauth2/default',
+  CLIENT_ID: process.env.SPA_CLIENT_ID || '{clientId}',
+  CLIENT_SECRET: '{clientSecret}',
```
jmelberg-okta: nit: We can remove this

packages/okta-vue/test/e2e/harness/package.json
vijetmahabaleshwar-okta: My bad. Will put it back to where it belongs.

.gitignore
```diff
@@ -5,3 +5,10 @@ npm-debug.log
 package-lock.json
 .envrc
 *.code-workspace
+lerna-debug.log
+packages/jwt-verifier/okta-oidc-tck-*
```
jmelberg-okta: Instead of placing these at the global `.gitignore`, they can be placed inside of their respective subdirectories.

robertjd: This can be made simpler by just doing this:

```
# Ignore TCK-related files in all folders
okta-oidc-tck*
target
```

CONTRIBUTING.md
```diff
@@ -41,3 +41,39 @@ If you do not see an exact emoji match, use the best matching emoji.
     Updates Contributing.md with new template
 
     Resolves: #1234
+
+
+## Running Tests locally
+
+You can run all the tests inside each of the packages from the project root.
```
robertjd: Let's add a bit more context here: you can run all the tests locally OR you can go into an individual folder an run just that packages tests.

PR #111, opened 2018-02-01
okta-vue - replace path when redirecting

### LINKED ISSUE #98: okta-vue - replace path instead of push it
When authentication is successful it redirects to a `callback` route in the app. And this route redirects to the page we were originally trying to access.

The problem is that the redirect is made with a push instead of a replace, that means that if the user hits the back button they will be back to the `callback` route which will fail.

```
Uncaught (in promise) n {name: "AuthSdkError", message: "Unable to parse a token from the url", errorCode: "INTERNAL", errorSummary: "Unable to parse a token from the url", errorLink: "INTERNAL", …}
```

Can we do the redirect with replace instead of push?

https://github.com/okta/okta-oidc-js/blob/f6f66aa3dfde922b4af5d80df655f07f7d37e5b0/packages/okta-vue/src/components/ImplicitCallback.vue#L6

gaastonsr:
Fixes #98.

### DIFF
--- a/packages/okta-vue/src/components/ImplicitCallback.vue
+++ b/packages/okta-vue/src/components/ImplicitCallback.vue
@@ -3,7 +3,7 @@ export default {
   name: 'ImplicitCallback',
   async beforeMount () {
     await this.$auth.handleAuthentication()
-    this.$router.push({
+    this.$router.replace({
       path: this.$auth.getFromUri()
     })
   },


jmelberg-okta:
Hey @gaastonsr! Can you rebase this against the branch `1.0.0-vue`? Since this is a breaking change, we're going to do a major version bump.

gaastonsr:
Hi @jmelberg-okta, I just rebased the branch against `1.0.0-vue` please let me know if it's ok.

PR #112, opened 2018-02-01
okta-vue - fix navigation guard arguments order

### LINKED ISSUE #110: okta-vue - beforeEach navigation guard arguments are backwards
I think I found the cause of one of the weird bugs I have in my application. Sometimes I enter and the homepage loads normally and then I click something and it redirects me to the login page. I think it's because the arguments to the navigation guard are backward.

https://github.com/okta/okta-oidc-js/blob/2364ff4af9a5f4f410f28f8a63ada2b6bcb653a0/packages/okta-vue/src/Auth.js#L53

In the source code we have `(from, to next)` and [in the documentation, it says](https://router.vuejs.org/en/advanced/navigation-guards.html) it should be `(to, from, next)`.

gaastonsr:
Fixes #110.

### DIFF
--- a/packages/okta-vue/src/Auth.js
+++ b/packages/okta-vue/src/Auth.js
@@ -50,7 +50,7 @@ function install (Vue, options) {
       return accessToken ? oktaAuth.token.getUserInfo(accessToken) : undefined
     },
     authRedirectGuard () {
-      return async (from, to, next) => {
+      return async (to, from, next) => {
         if (from.matched.some(record => record.meta.requiresAuth) && !(await this.isAuthenticated())) {
           localStorage.setItem('referrerPath', from.path || '/')
           this.loginRedirect()
--- a/packages/okta-vue/src/Auth.js
+++ b/packages/okta-vue/src/Auth.js
@@ -51,8 +51,8 @@ function install (Vue, options) {
     },
     authRedirectGuard () {
       return async (to, from, next) => {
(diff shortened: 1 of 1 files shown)


robertjd:
Thanks @gaastonsr , these certainly seem to be backwards [per docs](https://router.vuejs.org/en/advanced/navigation-guards.html).  It looks like we also need to change the login within this function to reference `to` rather than `from`.  I'm testing with the included testing, see instructions at the end of the README.

packages/okta-vue/src/Auth.js
```diff
@@ -50,7 +50,7 @@ function install (Vue, options) {
       return accessToken ? oktaAuth.token.getUserInfo(accessToken) : undefined
     },
     authRedirectGuard () {
-      return async (from, to, next) => {
+      return async (to, from, next) => {
         if (from.matched.some(record => record.meta.requiresAuth) && !(await this.isAuthenticated())) {
```
jmelberg-okta: In order to keep this functionality, you'll need to update this logic to use the new variable names.

`if (from.matched...) ` **needs to be** `if (to.matched...)`

`localStorage.setItem('referrerPath', from.path || '/')`  **needs to be** `localStorage.setItem('referrerPath', to.path || '/')`

You should be able to validate this locally by running `npm test` (after updating the file `test/e2e/harness/config/dev-env.js`.

gaastonsr:
Thanks, @jmelberg-okta. Do you think it would be possible to do a release soon?

PR #113, opened 2018-02-01
okta-vue - expose AuthJS to vue

### LINKED ISSUE #19: okta-react: Make Auth.js service available outside of React components

gaastonsr:
This one is closely related to #19. It's the equivalent of #97 but for Vue.

I couldn't run the tests though. I would appreciate it if somebody could verify it works or could guide me on how to do it.

Fixes #19.

### DIFF
--- a/packages/okta-vue/README.md
+++ b/packages/okta-vue/README.md
@@ -57,6 +57,26 @@ Vue.use(Auth, {
 
 ```
 
+#### Example with AuthJS object
+
+Provide an AuthJS object instead of the options above. This is helpful when integrating `okta-vue` with external libraries that need access to the tokens.
+
+```typescript
+// router/index.js
+
+import Auth from '@okta/okta-vue'
+
+const auth = new Auth.AuthJS({
+  issuer: 'https://{yourOktaDomain}.com/oauth2/default',
+  client_id: '{client_id}',
+  redirect_uri: 'http://localhost:{port}/implicit/callback',
+  scope: 'openid profile email'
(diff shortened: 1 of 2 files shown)


jmelberg-okta:
Hey @gaastonsr - I'm a little unsure about this change. While exposing this makes sense for flows that `okta-vue` currently doesn't support - it also deeply couples `okta-vue` to `okta-auth-js`.

What flows do we not support that you're using `okta-auth-js` for currently?

gaastonsr:
Right now we need to be inside a component to have access auth object. We would like to be able to access it from Vuex directly.

kevlened:
@gaastonsr #19 was resolved in #97 by creating an instance of `Auth` and passing it in. That solution makes sense here too, because it gets you access to the `Auth` methods and addresses @jmelberg-okta's concern of being coupled to `okta-auth-js`.

kevlened:
@gaastonsr Turns out that a similar solution won't work with the Vue implementation currently, but you can work around it by using the Vue global.

```javascript
Vue.use(Auth, {
  issuer: 'yours',
  client_id: 'yours',
  redirect_uri: 'yours',
  scope: 'openid',
});

Vue.prototype.$auth // Do what you need
```

PR #114, opened 2018-02-01
:bug: Fixes fromUri path + adds param during login and logout for navigation

### LINKED ISSUE #84: Document how to override the referrer to redirect to different page than login
Inspired by https://devforum.okta.com/t/implicit-callback-redirects-to-the-angular-component-and-then-to-homepage-automatically/960/6.

jmelberg-okta:
### Updates
- Automatically set `fromUri` value when the `AuthGuard` is triggered.
     - In the default case where no `onAuthRequired` is passed in, we pass the current `url` into the `loginRedirect` method. 
    - When `onAuthRequired` is passed in, the current url value must be set before executing the `onAuthRequired` method.
        - For example: `/protected` is blocked by the `OktaAuthGuard`, where a custom `onAuthRequired` logic navigates the user to `/login`.
- Update the default `OktaLoginRedirectComponent` to use either the default path `/` as the `fromUri` or the value provided by the `onAuthRequired` guard logic.
- Removes checking for the object type of scope (breaking change)
- Adds configurable url to redirect to after logout

### Resolves
#84 (Angular)

### DIFF
--- a/packages/okta-angular/README.md
+++ b/packages/okta-angular/README.md
@@ -101,13 +101,40 @@ Finally, import the `OktaAuthModule` into your `NgModule`, and instantiate it by
 export class MyAppModule { }
 ```
 
-### Reference
-**Configuration Options**
-  - `issuer` **(required)**: The OpenID Connect `issuer`
-  - `clientId` **(required)**: The OpenID Connect `client_id`
-  - `redirectUri` **(required)**: Where the callback is hosted
-  - `scope` *(optional)*: Reserved or custom claims to be returned in the tokens
-  - `onAuthRequired` *(optional)*: Accepts a callback to make a decision when authentication is required. If not supplied, `okta-angular` will redirect directly to Okta for authentication.
+## Use the Access Token
+When your users are authenticated, your Angular application has an access token that was issued by your Okta Authorization server. You can use this token to authenticate requests for resources on your server or API. As a hypothetical example, let's say you have an API that provides messages for a user. You could create a `MessageList` component that gets the access token and uses it to make an authenticated request to your server.
+
+Here is what the Angular component could look like for this hypothetical example:
+
+```typescript
+// messagelist.component.ts
+
+import { Component } from '@angular/core';
+import { Http, Headers, RequestOptions } from '@angular/http';
+import { OktaAuthService } from '@okta/okta-angular';
+import 'rxjs/Rx';
+
+@Component({
+  template: `
+    <div *ngIf="messages.length">
+      <li *ngFor="let message of messages">{{message.message}}</li>
+    </div>
+  `
+})
+export class MessageListComponent {
+  messages = [];
+  constructor(private oktaAuth: OktaAuthService, private http: Http) {
+    const headers = new Headers({ Authorization: 'Bearer ' + oktaAuth.getAccessToken().accessToken });
+    // Make request
+    this.http.get(
+      'http://localhost:{serverPort}/api/messages',
+      new RequestOptions({ headers: headers })
+    )
+    .map(res => res.json())
+    .subscribe((messages: Array<Object>) => messages.forEach(message => this.messages.push(message)));
+  }
+}
+```
 
 ### Using a custom login-page
 The `okta-angular` SDK supports the session token redirect flow for custom login pages. For more information, [see the basic Okta Sign-in Widget functionality](https://github.com/okta/okta-signin-widget#new-oktasigninconfig).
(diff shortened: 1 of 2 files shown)


packages/okta-angular/README.md
```diff
@@ -144,6 +171,52 @@ const oktaConfig = {
 };
 ```
 
+## Reference
+### `oktaAuth`
+`oktaAuth` is the top-most component of `okta-angular`. This is where most of the configuration is provided.
```
robertjd: I'm finding this section a bit confusing because `oktaAuth` is a property that we add to the scope, after you've called `OktaAuthModule.initAuth(oktaConfig)` with the configuration options that your mentioning.  This isn't really clear, currently it looks like oktaAuth is a method I should call with the configuration options?

jmelberg-okta: +1. This is a bit confusing.

packages/okta-angular/README.md
```diff
@@ -144,6 +171,52 @@ const oktaConfig = {
...
+If you have an Okta `sessionToken`, you can bypass the full-page redirect by passing in this token. This is recommended when using the [Okta Sign-In Widget](https://github.com/okta/okta-signin-widget). Simply pass in a `sessionToken` into the `loginRedirect` method follows:
+
+```typescript
+// Navigate to "/profile" on login success
+this.oktaAuth.loginRedirect('/profile', {
+  sessionToken: /* sessionToken */
+})
+```
+
+> Note: For information on obtaining a `sessionToken` using the [Okta Sign-In Widget](https://github.com/okta/okta-signin-widget), please see the [`renderEl()` example](https://github.com/okta/okta-signin-widget#rendereloptions-success-error).
```
robertjd: I know we're trying to wrap Auth JS and eventually take away the direct access, but we should put some kind of documentation in here about that these "additional params" are - even if it's just linking to the AuthJS method we're delegating to: https://github.com/okta/okta-auth-js#tokengetwithredirectoptions

packages/okta-angular/src/okta/okta.service.ts
```diff
@@ -147,21 +152,21 @@ export class OktaAuthService {
...
-    async logout() {
+    async logout(uri?: String) {
       this.oktaAuth.tokenManager.clear();
       await this.oktaAuth.signOut();
+      this.router.navigate([uri || '/']);
     }
 
     /**
      * Scrub scopes to ensure 'openid' is included
+     * @param scopes
```
robertjd: These JSDoc comments need more fidelity, for example:

```
@param {String} uri The location to navigate to after logout
```

packages/okta-angular/src/okta/okta.service.ts
```diff
@@ -97,8 +97,13 @@ export class OktaAuthService {
 
     /**
      * Launches the login redirect.
+     * @param fromUri
+     * @param additionalParams
      */
-    loginRedirect(additionalParams?: object) {
+    loginRedirect(fromUri: String, additionalParams?: object) {
```
robertjd: This signature change is also a breaking change

jmelberg-okta: Should this be pulled out as another PR?

packages/okta-angular/src/okta/okta.service.ts
robertjd: Or even better cuz this is typescript, just update the method:

```
scrubScope(scopes: string)
```

packages/okta-angular/README.md
```diff
@@ -144,6 +171,86 @@ const oktaConfig = {
...
+#### `oktaAuth.getAccessToken()`
+Returns the access token from storage (if it exists).
+
+#### `oktaAuth.getIdToken()`
+Returns the ID token from storage (if it exists).
+
+#### `oktaAuth.getUser()`
+Returns the result of the OpenID Connect `/userinfo` endpoint if an access token exists.
+
+#### `oktaAuth.handleAuthentication()`
```
robertjd: Also: this does not return and redirects you to the stored URL

packages/okta-angular/README.md
```diff
@@ -144,6 +171,86 @@ const oktaConfig = {
...
+Parses the tokens returned as hash fragments in the OAuth 2.0 Redirect URI.
+
+### `oktaAuth.logout(uri?)`
```
robertjd: Missing a 4th # on this header

packages/okta-angular/README.md
```diff
@@ -144,6 +171,86 @@ const oktaConfig = {
 };
 ```
 
+## Reference
```
robertjd: I think this reference section needs another pass.  Right now it appears to be serving multiple purposes:

- Show you all the available classes you can import
- Tells you about some configuration options (but doesn't tell you where they go or how to use them)
- Shows you how to do configure the final redirect URL for login and logout operations
- Documents some, but not all, of the public methods available on the `OktaAuthModule` service

The content in here is good, though I feel that it needs to be moved around to other sections, and that this "reference" section should be more focused on describing the `OktaAuthModule` and what it provides you.

packages/okta-angular/src/okta/okta.guard.ts
```diff
@@ -46,20 +46,17 @@ export class OktaAuthGuard implements CanActivate {
...
 
-    /** 
-     * Store the current path
-     */
-    this.oktaAuth.setFromUri(state.url);
-
     /**
      * Redirect to the given path or
      * perform the default Okta full-page redirect.
      */
```
robertjd: Does this comment block still make sense?

packages/okta-angular/src/okta/okta.service.ts
```diff
@@ -147,21 +152,21 @@ export class OktaAuthService {
...
      * Scrub scopes to ensure 'openid' is included
+     * @param scopes
      */
-    scrubScopes(scopes) {
+    scrubScopes(scopes: string) {
       if (!scopes) {
         return 'openid email';
-      } else {
-        // Make sure object is a string
-        scopes = Array.isArray(scopes) ? scopes.join(' ') : scopes
```
robertjd: Before merging, let's continue our conversation about conventional commits.  I think this breaking change needs to be it's own commit, and the change to loginRedirect() needs to be it's own commit.  There's a lot going on in this PR and someone scanning for breaking changes may have a hard time finding them.

jmelberg-okta:
Closing due to #125

PR #117, opened 2018-02-02
feat(angular): Add getUser to Angular

### LINKED ISSUE #107: getUser should return info from the `/userinfo` endpoint or decode `idToken`
if `accessToken` -> request `/userinfo` endpoint
if !`accessToken` -> decode `idToken` for claims

cmckni3:
Close #107

### DIFF
--- a/packages/okta-angular/src/okta/models/user-info.ts
+++ b/packages/okta-angular/src/okta/models/user-info.ts
@@ -1,0 +1,24 @@
+export class Address {
+  street_address: string;
+  locality: string;
+  region: string;
+  postal_code: string;
+  country: string;
+}
+
+export class UserInfo {
+  sub: string;
+  name: string;
+  nickname: string;
+  given_name: string;
+  middle_name: string;
+  family_name: string;
+  profile: string;
+  zoneinfo: string;
+  locale: string;
+  updated_at: number;
+  email: string;
+  email_verified: boolean;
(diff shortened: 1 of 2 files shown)


cmckni3:
We can talk here about my comment on #107. I am willing to help out with this code since I need it ASAP.

jmelberg-okta:
Hey @cmckni3! Thanks for all your work on the last couple of PRs. We're going to take a look at these as soon as we can.

Re: #107 
There are three primary use cases we need to support for `getUser`:
1. Requesting `id_token token` (default)
    - If there is an `accessToken` stored, make a call to the `/userinfo` endpoint to return all available user information based on the `scopes` requested.
2. Requesting `id_token`
    - `getUser` should decode the token and return the payload of the `jwt`
3. Requesting `token`
    - Same as 1.

As you mentioned, the claims in the `idToken` may become stale/expired. Common practice is to keep a short `idToken` expiration time - then rely on the underlying library [`AuthJS`](https://github.com/okta/okta-auth-js) to handle refreshing the tokens when needed.

cmckni3:
That makes sense @jmelberg-okta according to the Oauth specification. I mostly struggle with following the AuthJS code since none of the documentation is available in TypeScript.

cmckni3:
I worked around this in my current application by calling `loginRedirect`.

PR #118, opened 2018-02-07
Fixes #102

### LINKED ISSUE #102: Add support for failure redirect
PassportJS has a `failureRedirect` route which can be provided by client to show a custom error page.

The case i've is if you hit `authorization-code/callback?error=access_denied` if will just show `Unauthorized` and not an easy way to override this error page or show a page.

I propose, provided the clients ability to provide `failureRedirect` for passport similar to `successReturnToOrRedirect`.

ahmednasir91:
Added support for failureRedirect to show custom error page when authentication fails.


robertjd:
Thank you @ahmednasir91 , this should now be available in the 0.1.0 release of oidc-middleware

PR #119, opened 2018-02-07
Updates navigation paths on login/logout to make them configurable

jmelberg-okta:
BREAKING CHANGE

- *Optional*: Specify the redirect path when calling the `loginRedirect` method.
- *Optional*: Specify the redirect path when calling `logout`. (Defaults to `/`)

### DIFF
--- a/packages/okta-angular/src/okta/okta.service.ts
+++ b/packages/okta-angular/src/okta/okta.service.ts
@@ -99,8 +99,14 @@ export class OktaAuthService {
 
     /**
      * Launches the login redirect.
+     * @param fromUri
+     * @param additionalParams
      */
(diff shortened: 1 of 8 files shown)


packages/okta-angular/src/okta/okta.service.ts
```diff
@@ -97,8 +97,13 @@ export class OktaAuthService {
 
     /**
      * Launches the login redirect.
+     * @param fromUri
+     * @param additionalParams
      */
-    loginRedirect(additionalParams?: object) {
+    loginRedirect(fromUri: string, additionalParams?: object) {
```
cmckni3: `fromUri` should be optional.

PR #122, opened 2018-02-13
chore: Add client user agent

robertjd:
Adds our standard library user agent data

### DIFF
--- a/packages/oidc-middleware/test/unit/constructor.js
+++ b/packages/oidc-middleware/test/unit/constructor.js
@@ -1,0 +1,43 @@
+const expect = require('chai').expect;
+const nock = require('nock');
+const os = require('os');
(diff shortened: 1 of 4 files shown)


packages/oidc-middleware/test/unit/constructor.js
```diff
@@ -0,0 +1,43 @@
...
+        expect(userAgent).to.equal(expectedAgent);
+        done();
+      });
+      new ExpressOIDC({
+        client_id: 'foo',
+        client_secret: 'foo',
+        redirect_uri: 'foo',
+        issuer: 'http://foo'
+      }).on('error', () => {
+        // Because we mocking and not fulfilling the real response, the client will error
```
jmelberg-okta: we -> we're

PR #127, opened 2018-02-15
:memo: Updates README.md

inergy:
Updates README.md in the okta-react package with a correction in the Usage section (Callback should be ImplicitCallback)


robertjd:
Thank you @inergy !

Issue #123, opened 2018-02-14
okta-angular: Callback with Query String Parameters


th3n3wguy:
Problem:

- When using the `okta-angular` package, if a user's session is timed out, they are redirected to the Okta login page with a callback URL that contains query-string parameters. The problem is that the query string parameters are being converted using, what I assume to be, something like JavaScript's [encodeURIComponent](https://www.w3schools.com/jsref/jsref_encodeuricomponent.asp). All of the `?` and `&` and other symbols are converted, which breaks our UI application's URLs when the callback is handled within the `okta-angular` library to handle the redirect after a successful login.

Solution:

- No idea how to help here as I don't have the time to go through the code repository to find the exact lines of code that are handling this functionality. I apologize for not doing further code analysis, but hopefully you will be able to resolve this issue fairly quickly.

jmelberg-okta:
Hey @th3n3wguy! I want to get a bit more info on how you're using the `okta-angular` library.

Just for clarification, by default, the `loginRedirect` method performs the [Implicit Flow](https://developer.okta.com/authentication-guide/implementing-authentication/implicit), which is required to return the tokens back via a hash fragment. Is this the flow you're using?

Secondly, I'm assuming you've setup a redirect URI in Okta and your SPA to be formatted **with** query params. For example: `https://mysite.com/callback?q=someState`. After successful login, the tokens are appended to the callback URI similar to: `https://mysite.com/callback?q=someState#accessToken=ey...`.

Is the issue here that the specified callback includes the encoded query params, or is this only an issue when attempting to parse the tokens from the URL?

th3n3wguy:
What is happening for us is that if we call the `OktaAuthService.loginRedirect()` function from any page that contains query string parameters (e.g. `https://myui.sample.com/page1?q=test&base=super`), then we are redirected to the Okta login page and that workflow is functioning just fine, but as soon as the application redirects us back to the original page, the URL comes out looking like this:
- `https://myui.sample.com/page1%3Fq%3Dtest%26base%3Dsuper`

Then, Angular has no idea how to get the query string parameters because it doesn't recognize those HTML-encoded characters as query string parameters, so the page does not reload properly. In the meantime, we have it set up where pages automatically redirect back to our home page after re-authenticating with Okta, but this is not an ideal scenario for the long-term.

jmelberg-okta:
Hey @th3n3wguy.

[...]

PR #137, opened 2018-03-06
:bug: Fixes redirect back to protected page

jmelberg-okta:
BREAKING CHANGE

Fixes an issue where navigation back to a protected page was encoding the query parameters. Now, protected pages can include query params: `/protected?state=foo`

Resolves:
OKTA-161365
#123

### DIFF
--- a/packages/okta-angular/src/okta/okta.guard.ts
+++ b/packages/okta-angular/src/okta/okta.guard.ts
@@ -52,7 +52,7 @@ export class OktaAuthGuard implements CanActivate {
     /** 
      * Store the current path
      */
-    this.oktaAuth.setFromUri(state.url);
+    this.oktaAuth.setFromUri(route.url[0].path, route.queryParams);
 
     /**
      * Redirect to the given path or
--- a/packages/okta-angular/src/okta/okta.guard.ts
+++ b/packages/okta-angular/src/okta/okta.guard.ts
@@ -52,7 +52,8 @@ export class OktaAuthGuard implements CanActivate {
     /** 
      * Store the current path
      */
-    this.oktaAuth.setFromUri(route.url[0].path, route.queryParams);
+    const path = state.url.split(/[?#]/)[0];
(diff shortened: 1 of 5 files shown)


packages/okta-angular/src/okta/okta.guard.ts
```diff
@@ -52,7 +52,7 @@ export class OktaAuthGuard implements CanActivate {
     /** 
      * Store the current path
      */
-    this.oktaAuth.setFromUri(state.url);
+    this.oktaAuth.setFromUri(route.url[0].path, route.queryParams);
```
robertjd: This won't work for paths like `foo/bar`

packages/okta-angular/src/okta/okta.guard.ts
```diff
@@ -52,7 +52,12 @@ export class OktaAuthGuard implements CanActivate {
     /** 
      * Store the current path
      */
-    this.oktaAuth.setFromUri(state.url);
+    let path: Array<string> = [];
+    for (var i in route.url) {
+      path.push(route.url[i].path);
+    }
+
+    this.oktaAuth.setFromUri(path.join('/'), route.queryParams);
```
th3n3wguy: Thought I would leave this here, but here is another way you could do this (trying to be helpful any way I can):
```javascript
const path: string = Object.keys(route.url)
  .map((val: UrlSegment) => val.path)
  .join('/');

this.oktaAuth.setFromUri(path, route.queryParams);
```

packages/okta-angular/src/okta/okta.guard.ts
```diff
@@ -52,7 +52,8 @@ export class OktaAuthGuard implements CanActivate {
     /** 
      * Store the current path
      */
-    this.oktaAuth.setFromUri(state.url);
+    const path = state.url.split(/[?#]/)[0];
```
jmelberg-okta: As long as routes are configured as **child routes** (Angular's design decision) this works by removing the `#` and `?` from the URL.

PR #141, opened 2018-03-08
Updates to RN client

jmelberg-okta:
Updates the proposed React Native library by adding the following commits:
1. 🐛 Fixes an `default` import issue 
2. Updates the `README` to provide more detailed instructions
3. Remove nested `/packages` directory (duplicated code)
4. Adds license headers and `LICENSE` file to the root of the library

### Testing
```
cd packages/okta-react-native
npm install
npm test
```
**TODO:** Once merged into the active `react-native-client` branch, add test instructions for Travis


packages/okta-react-native/LICENSE
```diff
@@ -0,0 +1,187 @@
+The Okta software accompanied by this notice is provided pursuant to the
+following terms:
+
+Copyright © 2015-2018, Okta, Inc.
```
robertjd: If this is the first year that we've written the code in this library, then we should have just 2018 for now.

Issue #109, opened 2018-01-31
Make `responseType` configurable as a String


jmelberg-okta:
Support `id_token`, `token`, `id_token token`, etc `responseType`

```
config = {
  ...
  responseType: 'id_token token'
}
```

jmelberg-okta:
Resolved with #157 and #167

PR #143, opened 2018-03-09
:seedling: Make responseType configurable

jmelberg-okta:
### Description
Allows the OAuth 2.0 `response_type` to be configurable for implicit flows.

**Important:** `code` isn't explicitly supported yet.

Resolves: #109


packages/okta-angular/test/e2e/harness/src/app/app.module.ts
```diff
@@ -48,7 +48,7 @@ const appRoutes: Routes = [
   },
   {
     path: 'sessionToken-login',
-    component: SessionTokenLogin
+    component: SessionTokenLoginComponent
```
robertjd: This looks like a breaking change?

packages/okta-angular/test/e2e/harness/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import { environment } from './../environments/environment';
+
+import {
+  OktaAuthGuard,
+  OktaAuthModule,
+  OktaCallbackComponent,
+  OktaLoginRedirectComponent
+} from '@okta/okta-angular';
+
+describe('Unit Tests', () => {
```
robertjd: These spec files should be siblings of the source files, it's odd that we're burying unit testing within an e2e test.  We should discuss this, we need to decide if we're going to fix this now or later.  I also need a place to put unit testing for the user agent work

packages/okta-angular/src/okta/okta.config.ts
```diff
@@ -18,6 +18,7 @@ export interface OktaConfig {
   clientId?: string;
   scope?: string;
   onAuthRequired?: Function;
+  responseType?: string;
```
robertjd: Should get some README love

packages/okta-angular/test/e2e/harness/src/app/app.module.ts
jmelberg-okta: This is used only in our test suite. We aren't exposing any of this via our SDK, so we aren't introducing a breaking change.

packages/okta-angular/README.md
```diff
@@ -37,6 +37,7 @@ The `OktaAuthModule` is the initializer for your OpenID Connect client configura
   - `clientId` **(required)**: The OpenID Connect `client_id`
   - `redirectUri` **(required)**: Where the callback is hosted
   - `scope` *(optional)*: Reserved for custom claims to be returned in the tokens
+  - `responseType` *(optional)*: Desired token grant types
```
robertjd: I feel like we should be explicitly mentioning that these are string values.

packages/okta-angular/test/e2e/harness/src/app/app.component.spec.ts
robertjd: As discussed offline: this is going to be significant work and duplication of test boilerplate, so we won't do this.

packages/okta-angular/README.md
jmelberg-okta: I thought about noting that in this section, but seems weird to only mention the types on this configuration value. When this value is evaluated, typescript will error due to a type difference.

robertjd: Aw, types 💘

PR #144, opened 2018-03-09
Testing failure on travis

vijetmahabaleshwar-okta:


scripts/runTests.sh
```diff
@@ -1,8 +1,20 @@
...
+    if [ $? -ne 0 ]; then
+        STATUS=1
+        FAILED_PACKAGE=$package
+    fi
     cd ../..
 done
+
+if [ ! -z $FAILED_PACKAGE ]; then
+    echo $FAILED_PACKAGE
+fi
```
jmelberg-okta: Instead of relying on the var `STATUS`, we can exit inside of this conditional once we `echo $FAILED_PACKAGE`

robertjd: Multiple packages could fail.  The goal of this fix should be to exit non-zero if any of them fail.  The individual tests write error information to output so we already know which ones failed.

jmelberg-okta: Makes sense. Just to clarify - are you saying we should have a hard fail once **any** package fails?

For example, if `okta-angular` fails - exit immediately regardless of the status of other tests.

robertjd: We should probably exit immediately, since the total test run time is already getting close to 10 minutes

vijetmahabaleshwar-okta: Updated the script to exit immediately on failure.

PR #147, opened 2018-03-14
:seedling: Adds getUser method

jmelberg-okta:
Adds a helper method to return a payload of user information, based on the available tokens.

This method considers the token substitution attack vector mentioned [here](http://openid.net/specs/openid-connect-core-1_0.html#TokenSubstitution), using the recommendations outlined in [OpenID Connect UserInfo response](http://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse).

- [x] If there is an `accessToken` and `idToken` ➡️ Return the most up-to-date record from the `/userinfo` endpoint.
- [x] If there is **no** `accessToken`, but an `idToken` is present ➡️ Decode the `idToken` payload.

Resolves: #107

### DIFF
--- a/packages/okta-angular/src/okta/okta.service.ts
+++ b/packages/okta-angular/src/okta/okta.service.ts
@@ -89,6 +89,24 @@ export class OktaAuthService {
     }
 
     /**
+     * Returns user claims from the /userinfo endpoint if an
+     * accessToken is provided or parses the available idToken.
+     */
+    async getUser() {
+      const accessToken = this.oktaAuth.tokenManager.get('accessToken');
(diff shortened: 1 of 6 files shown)


packages/okta-angular/src/okta/okta.service.ts
```diff
@@ -88,6 +88,24 @@ export class OktaAuthService {
       return this.oktaAuth.tokenManager.get('idToken');
     }
 
+    /**
+     * Returns user claims from the /userinfo endpoint if an
+     * accessToken is provided or parses the available idToken.
+     */
+    async getUser() {
```
robertjd: Needs README <3

PR #148, opened 2018-03-14
:seedling: Updates getUser method

jmelberg-okta:
Adds a helper method to return a payload of user information, based on the available tokens.

This method considers the token substitution attack vector mentioned [here](http://openid.net/specs/openid-connect-core-1_0.html#TokenSubstitution), using the recommendations outlined in [OpenID Connect UserInfo response](http://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse).

- [x] If there is an `accessToken` and `idToken` ➡️ Return the most up-to-date record from the `/userinfo` endpoint.
- [x] If there is **no** `accessToken`, but an `idToken` is present ➡️ Decode the `idToken` payload.

Resolves: #107

### DIFF
--- a/packages/okta-vue/src/Auth.js
+++ b/packages/okta-vue/src/Auth.js
@@ -47,7 +47,16 @@ function install (Vue, options) {
     },
     async getUser () {
       const accessToken = oktaAuth.tokenManager.get('accessToken')
-      return accessToken ? oktaAuth.token.getUserInfo(accessToken) : undefined
+      const idToken = oktaAuth.tokenManager.get('idToken')
+      if (accessToken && idToken) {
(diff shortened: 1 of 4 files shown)


jmelberg-okta:
@robertjd This library hardcodes the `response_type` to `id_token token`, so in most cases we can guarantee the user has an access token and id token.

It is possible that someone could have pulled out the `authJS` object and rewrote our [`loginRedirect `](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-vue/src/Auth.js#L14) method, but they would also need to modify the [`handleAuthetication`](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-vue/src/Auth.js#L28) method to account for a single token. I find this *very* unlikely and outside of the intended usage of this lib.

PR #149, opened 2018-03-14
:bug: Fix flaky E2E angular tests

jmelberg-okta:
Fixes a flaky test where we were checking for `/login` in the URL path instead of the domain name change.

### DIFF
--- a/packages/okta-angular/test/e2e/harness/e2e/page-objects/okta-signin.po.ts
+++ b/packages/okta-angular/test/e2e/harness/e2e/page-objects/okta-signin.po.ts
@@ -18,8 +18,10 @@ export class OktaSignInPage {
       browser.waitForAngularEnabled(false);
     }
 
-    waitUntilVisible() {
-      browser.wait(ExpectedConditions.urlContains('/login'), 5000);
+    waitUntilVisible(issuer: string) {
+      // Check for a redirect to the Okta Hosted Login Page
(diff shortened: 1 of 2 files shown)


packages/okta-angular/test/e2e/harness/e2e/page-objects/okta-signin.po.ts
```diff
@@ -18,8 +18,10 @@ export class OktaSignInPage {
       browser.waitForAngularEnabled(false);
     }
 
-    waitUntilVisible() {
-      browser.wait(ExpectedConditions.urlContains('/login'), 5000);
+    waitUntilVisible(issuer: string) {
+      // Check for a redirect to the Okta Hosted Login Page
```
vijetmahabaleshwar-okta: Do we need the whole authorize url to be constructed here?
Can't we just wait if url contains `/oauth2/default/v1/authorize`?

jmelberg-okta: When we redirect to Okta for login, we modify the current URL by extracting the `/authorize` query params into a stored reference (`okta_key`).

The only consistent URL is the `baseURL` of our org.

vijetmahabaleshwar-okta: In that case should we rename `authorizeUrl` to `baseUrl`?

PR #151, opened 2018-03-19
Set user agent for Angular, React, Vue libs - plus Jest

robertjd:
Uses the new property in AuthJS to append the framework lib to the user agent

Also adds unit testing via Jest for React, Vue

Angular not yet tested, will add a "unit" test to app.component.spec.ts after #143 lands in master

### DIFF
--- a/packages/okta-vue/package.json
+++ b/packages/okta-vue/package.json
@@ -13,7 +13,7 @@
     "pretest": "npm run build && npm run build:harness",
     "prepublish": "npm run build",
     "jest": "jest src/",
-    "test": "npm run jest && npm run --prefix test/e2e/harness/ test",
+    "test": "npm run lint && npm run jest && npm run --prefix test/e2e/harness/ test",
     "start": "npm run --prefix test/e2e/harness/ start",
     "build": "rimraf dist/ && cross-env NODE_ENV=production webpack --config webpack.config.js --output-library-target=umd -p",
     "build:harness": "npm --prefix test/e2e/harness install",
--- a/packages/okta-vue/package.json
+++ b/packages/okta-vue/package.json
@@ -8,14 +8,32 @@
     "src"
   ],
   "scripts": {
+    "prebuild": "npm run build:package-info",
     "prestart": "npm run build",
     "pretest": "npm run build && npm run build:harness",
     "prepublish": "npm run build",
-    "test": "npm run --prefix test/e2e/harness/ test",
+    "jest": "jest src/",
+    "test": "npm run jest && npm run --prefix test/e2e/harness/ test",
     "start": "npm run --prefix test/e2e/harness/ start",
     "build": "rimraf dist/ && cross-env NODE_ENV=production webpack --config webpack.config.js --output-library-target=umd -p",
     "build:harness": "npm --prefix test/e2e/harness install",
+    "build:package-info": "node ../../util/write-package-info.js . src/packageInfo.js",
     "lint": "eslint --ext .js,.vue src"
+  },
+  "jest": {
+    "moduleFileExtensions": [
+      "js",
+      "json",
+      "vue"
+    ],
+    "transform": {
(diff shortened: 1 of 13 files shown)


packages/okta-vue/package.json
```diff
@@ -8,15 +8,33 @@
     "src"
   ],
   "scripts": {
+    "prebuild": "npm run build:package-info",
     "prestart": "npm run build",
     "pretest": "npm run build && npm run build:harness",
     "prepublish": "npm run build",
-    "test": "npm run --prefix test/e2e/harness/ test",
+    "jest": "jest src/",
+    "test": "npm run jest && npm run --prefix test/e2e/harness/ test",
```
jmelberg-okta: I'm leaning towards thinking we should prepend `npm run lint` to this command. That way we can start enforcing the `eslint` style.

util/write-package-info.js
```diff
@@ -0,0 +1,17 @@
...
+const workingDirectory = process.argv[2];
+const destinationFile = process.argv[3];
+const packageJsonPath = path.join(process.cwd(), workingDirectory, 'package.json')
+const configDest = path.join(process.cwd(), workingDirectory, destinationFile);
+const packageJson = require(packageJsonPath);
+const packageInfo = {
+  name: packageJson.name,
+  version: packageJson.version
+};
+const output = 'export default ' + JSON.stringify(packageInfo, null, 2) + ';'
```
jmelberg-okta: nit: There is no `;` in our `okta-vue/src/*.js` files.

util/write-package-info.js
```diff
@@ -0,0 +1,17 @@
+const fs = require('fs');
+const path = require('path');
+
+const workingDirectory = process.argv[2];
+const destinationFile = process.argv[3];
+const packageJsonPath = path.join(process.cwd(), workingDirectory, 'package.json')
```
jmelberg-okta: Missing semicolon

packages/okta-vue/package.json
```diff
@@ -8,15 +8,33 @@
...
     "start": "npm run --prefix test/e2e/harness/ start",
     "build": "rimraf dist/ && cross-env NODE_ENV=production webpack --config webpack.config.js --output-library-target=umd -p",
     "build:harness": "npm --prefix test/e2e/harness install",
+    "build:package-info": "node ../../util/write-package-info.js . src/packageInfo.js",
```
jmelberg-okta: We should consider using [`lerna run`](https://github.com/lerna/lerna#run) to handle this for us.

robertjd: This task needs to be integrated into the build pipeline of each package, and ran locally while working on that specific package, so I'm not sure how lerna would help here?

util/write-package-info.js
robertjd: This util script is at the root, so not part of lerna.  I can put together a PR to add eslint to the root, but we'd have to do testing to make sure it doesn't break what we just fixed for windows.

util/write-package-info.js
```diff
@@ -0,0 +1,17 @@
...
+const workingDirectory = process.argv[2];
+const destinationFile = process.argv[3];
+const packageJsonPath = path.join(process.cwd(), workingDirectory, 'package.json');
+const configDest = path.join(process.cwd(), workingDirectory, destinationFile);
+const packageJson = require(packageJsonPath);
+const packageInfo = {
+  name: packageJson.name,
+  version: packageJson.version
+};
+const output = 'export default ' + JSON.stringify(packageInfo, null, 2) + ';';
```
jmelberg-okta: As noted earlier, the trailing `+ ';'` will result in eslint issues in our Vue package. Can we make this conditional?

robertjd: I excluded this file using the eslintignore in the vue package

jmelberg-okta: Really? I'm seeing this after a clean install:
screen shot 2018-03-19 at 9 23 04 am

PR #152, opened 2018-03-19
Publish

robertjd:
# Changelog

### Features

e02a0ff2386acc2e853142ad608638850e109fa2 - Adds getUser method (#147)

### Other

fc794ef14b368973bf53ff1749f45ade9742361b - Set user agent on AuthJS

### Other

cdc04c74fbfc60a9d1fec390abfb7fbf10c36d50  - Update dependencies

### Other

fc794ef14b368973bf53ff1749f45ade9742361b - Set user agent on AuthJS

### Features

8b5200b740c6cdc914c98d324a7128bb9f4ae05c  - Adds getUser helper method (#148)

### Fixes

2c4dd6dc31cce97209dd872808e0e3e40963147c - fix navigation guard arguments order (#112)

### Other

fc794ef14b368973bf53ff1749f45ade9742361b - Set user agent on AuthJS

### DIFF
--- a/packages/okta-react/package.json
+++ b/packages/okta-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@okta/okta-react",
-  "version": "0.0.12",
+  "version": "0.0.13",
   "description": "React support for Okta",
   "main": "./dist/index.js",
   "scripts": {
--- a/packages/oidc-middleware/package.json
+++ b/packages/oidc-middleware/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@okta/oidc-middleware",
-  "version": "0.1.0",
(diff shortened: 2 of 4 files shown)


packages/okta-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@okta/okta-react",
-  "version": "0.0.12",
+  "version": "0.0.13",
```
jmelberg-okta: Any reason we aren't bumping this to `0.1.0` like the others?

robertjd: I don't consider the user agent change to be a feature add or functionality add.  It's not technically a bug fix but still feels more like a patch and a minor change.

PR #153, opened 2018-03-19
:seedling: Make response_type configurable

jmelberg-okta:
### Description
Allows the OAuth 2.0 `response_type` to be configurable for implicit flows.

**Important:** `code` isn't explicitly supported yet.

Resolves: #109

### DIFF
--- a/packages/okta-vue/src/Auth.spec.js
+++ b/packages/okta-vue/src/Auth.spec.js
@@ -2,16 +2,19 @@ import AuthJS from '@okta/okta-auth-js'
 import { createLocalVue } from '@vue/test-utils'
 import { default as Auth } from './Auth'
 
-const mockAuthJsInstance = {userAgent: 'foo'}
 const pkg = require('../package.json')
-
 jest.mock('@okta/okta-auth-js')
+
+const mockAuthJsInstance = {
+  userAgent: 'foo',
+  token: {
+    getWithRedirect: jest.fn()
+  }
+}
 
 AuthJS.mockImplementation(() => {
   return mockAuthJsInstance
 })
-
-const localVue = createLocalVue()
 
 describe('Auth', () => {
   test('is a Vue plugin', () => {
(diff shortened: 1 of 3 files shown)


packages/okta-vue/src/Auth.spec.js
```diff
@@ -11,17 +11,28 @@ AuthJS.mockImplementation(() => {
...
     expect(mockAuthJsInstance.userAgent).toMatch(expectedUserAgent)
   })
+  test('sets the context for the $auth plugin', () => {
+    localVue.use(Auth, {
+      issuer: '1',
+      client_id: '2',
+      redirect_uri: '3',
+      response_type: 'a b'
+    })
+    expect(localVue.prototype.$auth).toBeDefined()
```
jmelberg-okta: We don't expose the configuration or the `AuthJS` object in Vue, so there is no good way to check this feat.

Instead, we're ensuring the object still will be created.

robertjd: Using mocking you can still test the desired side effect: that `oktaAuth.token.getWithRedirect` is called with the right values, given the input options given to the Auth plugin constructor.  Since there is already a mock of the internal authJs instance in this test file, you can use `jest.fn()` to add a mock version of `token.getWithRedirect()` to it and then use the [`mock.calls` property](https://facebook.github.io/jest/docs/en/mock-function-api.html#mockfnmockcalls) to see that it was called with the expected response_type

packages/okta-vue/src/Auth.spec.js
```diff
@@ -17,11 +22,26 @@ describe('Auth', () => {
...
+      client_id: '2',
+      redirect_uri: '3',
+      scope: 'foo bar',
+      response_type: 'token'
+    })
+    localVue.prototype.$auth.loginRedirect()
+    const mockCallValues = mockAuthJsInstance.token.getWithRedirect.mock.calls[0][0]
+    expect(mockCallValues.responseType).toEqual(expect.arrayContaining(['token']))
+    expect(mockCallValues.scopes).toEqual(expect.arrayContaining(['foo', 'bar']))
+  })
```
robertjd: 👍 awesome!  I think we should also add another test that asserts that `response_type` will default to `['id_token', 'token']` if not overridden by config, because this is a library decision that we are explicitly making.

PR #159, opened 2018-03-22
:seedling: Adds ability to override redirect path on login

jmelberg-okta:
- Specify the redirect path when calling the `loginRedirect` method.

### DIFF
--- a/packages/okta-vue/test/e2e/harness/src/App.vue
+++ b/packages/okta-vue/test/e2e/harness/src/App.vue
@@ -2,7 +2,7 @@
   <div id="app">
     <router-link to="/" tag="button" id='home-button'> Home </router-link>
     <button v-if='authenticated' v-on:click='logout' id='logout-button'> Logout </button>
-    <button v-else v-on:click='$auth.loginRedirect' id='login-button'> Login </button>
+    <button v-else v-on:click='$auth.loginRedirect("/")' id='login-button'> Login </button>
     <router-link to="/protected" tag="button"> Protected </router-link>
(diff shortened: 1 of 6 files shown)


packages/okta-vue/test/e2e/harness/src/components/SessionTokenLogin.vue
```diff
@@ -28,7 +28,7 @@ export default {
         password: this.password
       })
       .then(res =>
-        this.$auth.loginRedirect({
+        this.$auth.loginRedirect('/protected', {
```
jmelberg-okta: To validate we are navigating to a custom path, we're passing in `/protected` here. Previously, we would hit `/`.

packages/okta-vue/test/e2e/harness/src/App.vue
```diff
@@ -2,7 +2,7 @@
   <div id="app">
     <router-link to="/" tag="button" id='home-button'> Home </router-link>
     <button v-if='authenticated' v-on:click='logout' id='logout-button'> Logout </button>
-    <button v-else v-on:click='$auth.loginRedirect' id='login-button'> Login </button>
+    <button v-else v-on:click='$auth.loginRedirect("/")' id='login-button'> Login </button>
```
jmelberg-okta: This is the primary reason for the breaking change.

Fun fact - If no params are passed in, Vue will set the first parameter as a `MouseEvent` object.

robertjd: o_0

PR #165, opened 2018-03-27
Make oktaAuth.isAuthenticated() async

robertjd:
This needs to be done because the underlying token getters will become async

As a result this method can no longer be directly bound to for value updates.  As such, implement an observable to make authentication state changes observable in components

Should be reviewed with https://github.com/okta/samples-js-angular/pull/13

### DIFF
--- a/packages/okta-angular/README.md
+++ b/packages/okta-angular/README.md
@@ -3,7 +3,8 @@
 An Angular (4+) wrapper around [Okta Auth JS](https://github.com/okta/okta-auth-js), that builds on top of Okta's [OpenID Connect API](https://developer.okta.com/docs/api/resources/oidc.html).
 
 This library currently supports:
-  - [OAuth 2.0 Implicit Flow](https://tools.ietf.org/html/rfc6749#section-1.3.2)
+
+- [OAuth 2.0 Implicit Flow](https://tools.ietf.org/html/rfc6749#section-1.3.2)
 
 ## Getting Started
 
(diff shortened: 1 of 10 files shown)


packages/okta-angular/README.md
```diff
@@ -231,26 +242,26 @@ this.oktaAuth.loginRedirect('/profile', {
 
 #### `oktaAuth.isAuthenticated()`
 
-Returns `true` if there is a valid access token or ID token.
+Returns a promise that returns `true` if there is a valid access token and ID token.
```
jmelberg-okta: This needs to be changed back to an `or`.

We should be gating authentication based on the presence of an `idToken` (first) or `accessToken`

packages/okta-angular/src/okta/models/user-claims.ts
```diff
@@ -0,0 +1,29 @@
...
+  iss?: string;
+  jti?: string;
+  locale?: string;
+  name?: string;
+  nonce?: string;
+  preferred_username?: string;
+  sub: string;
+  updated_at: Number;
+  ver: Number;
+  zoneinfo: string;
```
jmelberg-okta: `updated_at`, `ver`, and `zoneinfo` may not be included in the `/userinfo` response, so we should make these optional. See the [Standard Claims](http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims).

packages/okta-angular/src/okta/services/okta.service.ts
```diff
@@ -72,10 +81,16 @@ export class OktaAuthService {
...
     /**
-     * Checks if there is a current accessToken in the TokenManager.
+     * Checks if there is an access token and id token
      */
-    isAuthenticated(): boolean {
-      return !!this.oktaAuth.tokenManager.get('accessToken');
+    async isAuthenticated(): Promise<boolean> {
+      const accessToken = await this.getAccessToken()
+      const idToken = await this.getIdToken()
+      return !!(accessToken && idToken);
```
jmelberg-okta: This **should** be an **if**:
`return !!(accessToken || idToken)`;

packages/okta-angular/src/okta/models/user-claims.ts
```diff
@@ -0,0 +1,29 @@
...
+  jti?: string;
+  locale?: string;
+  name?: string;
+  nonce?: string;
+  preferred_username?: string;
+  sub: string;
+  updated_at?: Number;
+  ver?: Number;
+  zoneinfo?: string;
+  [propName: string]: any;  // For custom claims that may be configured by the org admin
```
jmelberg-okta: Should this be optional as well?

robertjd: This inherently is optional, see "Excess Property Checks" here https://www.typescriptlang.org/docs/handbook/interfaces.html

robertjd:
LGTM 👍

PR #173, opened 2018-03-31
Added the ability to configure the token manager in Angular

bkoliba:
This proposed change will add the ability to configure the okta-auth-js token manager in the Angular package. Given that okta-auth-js already has this ability to perform this configuration with all the proper default setting and error handling shouldn't have any breaking changes ([code reference](https://github.com/okta/okta-auth-js/blob/c4906a2eefc6b492b6cccb28ccc791c116c5e768/lib/TokenManager.js#L132-L179)). Also added a simple interface (OktaTokenManagerConfig) to highlight the current options available for configuring okta-auth-js token manager which used as an optional property in OktaConfig.

### DIFF
--- a/packages/okta-angular/package.json
+++ b/packages/okta-angular/package.json
@@ -37,20 +37,21 @@
     "@okta/okta-auth-js": "^1.14.0"
   },
   "devDependencies": {
-    "@angular/common": "^4.4.3",
-    "@angular/compiler": "^4.4.3",
-    "@angular/compiler-cli": "^4.4.3",
-    "@angular/core": "^4.4.3",
-    "@angular/platform-browser": "^4.4.3",
-    "@angular/router": "^4.4.3",
-    "rxjs": "^5.4.3",
-    "typescript": "^2.5.2"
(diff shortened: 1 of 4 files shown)


bkoliba:
First off thanks for creating this package it saved our team a lot effort for adding authentication in our new app(currently using this package with these changes and referencing the package file directly). Here is the reason for needing these changes. Presently working with a client who just got set up with Okta last year and built their wrapper around okta-auth-js. I feel this change to the angular package will fulfill their current needs and will work for the new applications without us managing our package.  The only thing it is missing currently for us is the ability to set up the token manager storage as sessionStorage. This is a security concern of theirs and doesn't mine reauthenticating after the tab is closed since on premises they have SSO configured with Okta.

Thanks for any feedback or considerations.

manueltanzi-okta:
@bkoliba  We just released:
- @okta/okta-angular v1.1.0
- @okta/okta-react v1.2.0
- @okta/okta-vue v1.1.0

Which add the possibility to configure TokenManager parameters, check it out!

PR #177, opened 2018-04-05
Switch e2e tests to run on port 8080

robertjd:
This makes it easier to re-use sample configuration that may already be in your environment


robertjd:
@vijetmahabaleshwar-okta for this to work I assume well need to update the org that is used by the travis environment, to add 8080 as an allowed redirect

PR #181, opened 2018-04-09
okta-react: Adds function refreshTokens to Auth.js

### LINKED ISSUE #180: okta-react: Add function refreshTokens to Auth.js to obtaining new tokens without redirect
Add a new function to [Auth.js](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-react/src/Auth.js) so that one can obtain new tokens in the background of an React application without redirecting the page.

This is similar to [auth.redirect({sessionToken})](https://github.com/okta/okta-oidc-js/tree/master/packages/okta-react#authredirectsessiontoken). But without a redirect, so that it will result in a much better user experience.

Implementation can be done by using the function [token.getWithoutPrompt(oauthOptions)](https://github.com/okta/okta-auth-js#tokengetwithoutpromptoauthoptions) of the underlying `_oktaAuth` client. 

I'd suggest to call this new function `refreshTokens`, because `get...Token` is already used to access the `id_token` and `access_token` locally from the browser.

mschaeffner:
to obtain new tokens without a redirect

Fixes #180

### DIFF
--- a/packages/okta-react/README.md
+++ b/packages/okta-react/README.md
@@ -296,6 +296,10 @@ auth.redirect({
 });
 ```
 
+#### `auth.refreshTokens({sessionToken})`
+
+Refreshes the `id_token` and `access_token` without prompting the user to log in again. Adds the newly obtained tokens to the underlying `tokenManager` and also returns tokens in promise. Parameter `sessionToken` is optional.
+
 #### `auth.handleAuthentication()`
 
 Parses tokens from the url and stores them.
--- a/packages/okta-react/src/Auth.js
+++ b/packages/okta-react/src/Auth.js
@@ -39,8 +39,12 @@ export default class Auth {
   }
 
   async handleAuthentication() {
-    let tokens = await this._oktaAuth.token.parseFromUrl();
-    tokens = Array.isArray(tokens) ? tokens : [tokens];
+    const tokenOrTokens = await this._oktaAuth.token.parseFromUrl();
+    this.addTokensToManager(tokenOrTokens);
+  }
+
+  addTokensToManager(tokenOrTokens) {
+    const tokens = Array.isArray(tokenOrTokens) ? tokenOrTokens : [tokenOrTokens];
     for (let token of tokens) {
       if (token.idToken) {
         this._oktaAuth.tokenManager.add('idToken', token);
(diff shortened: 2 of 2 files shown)


sabs231:
Hi, @mschaeffner this is a great PR, I need this functionality. Could you update the documentation on where is the best place to use it? If possible do you an ETA on when this feature will be available?

robertjd:
Hi @mschaeffner @sabs231 - help me understand the story that this is useful for? I'm curious because the underlying Auth JS library does try to do the silent refresh for you.  Unfortunatley we do have a race condition where it will fail sometimes (see https://github.com/okta/okta-auth-js/issues/92#issuecomment-374820247 , we're going to work on this soon). I'm  just curious to know if that's the issue you're running into, of if there is another use case where you need to manually refresh the tokens?

mschaeffner:
I already had a look into the the underlying `okta-auth-js`, especially the file `TokenManager.js`. But it seems like that a token gets refreshed after it expires.
https://github.com/okta/okta-auth-js/blob/1e3f1003be7fd94476c457ee1dd1a319157ce6b1/lib/TokenManager.js#L44

This leads to a situation, that there is a time frame in which my application doesn't have a valid token. Then I will get a 401 response when I try to fetch data from my API, because the token already expired.  A 401 response will redirect the user to the login screen. And the requirement for my application is that this must never ever happen. The application must run for months without forcing the user to re-authenticate.

So I opened this pull request, that as a developer I have more control over the token refresh logic and I can better integrate it into my application logic and requirements.

@robertjd, if you think the logic for a token refresh should be in the underlying Auth JS lib, that's fine with me. But then we need a way to configure it here in `okta-oidc-js`. E.g. in React a property of the `Security` class. So that you can specify: "Auto-refresh the token X seconds **before** it expires".

robertjd:
Thanks @mschaeffner , knowing that helps.  The change I want to see it that `tokenManager.get()` will become an async call, and internally it will attempt to validate the token after it's pulled from storage.  If it sees that it's expired it will attempt to refresh it for you before calling back to you.  The idea is that `tokenManager.get()` should always return the desired token unless it truly can't be refreshed (i.e. the okta session has been cleared).  If we made this change, would you still want a proactive refresh X seconds before the token expires?  CC @jmelberg-okta since we've been talking about this.

mschaeffner:
If you implement `tokenManager.get()` the way you described above, then I would definitely use it, because it solves the problem. Of course, you then have to do small changes in the higher level libraries (react, angular, vue).

bbicodyduval:
@robertjd We're also experiencing this issue, and your approach feels like the right one.  Would love to see this prioritized.

robertjd:
Hi all, we've implemented the changes that were discussed in this thread.  Please update to `okta-react@1.0.3`, which will use the new version of AuthJS.  Let us know how it goes!

PR #194, opened 2018-04-19
:memo: Update README.md

gambtho:
Incorrect parameter used to determine if Auth is required when routing 

per https://github.com/okta/okta-oidc-js/blob/d9326b11c5b9be2c67830e49320e236d89854b94/packages/okta-vue/src/Auth.js#L68 would assume this should be "to" instead of "from"


jmelberg-okta:
Closing with respect to #195. Thanks for the PR @gambtho!

PR #175, opened 2018-04-02
:seedling: Retains search and hash on ImplicitCallback redirect

pallavi:
Saves full referrer location object on login
Redirects to referrer location object from ImplicitCallback

Resolves: #174

Issue #174, opened 2018-04-02
ImplicitCallback doesn't retain query params or hash in the URL


pallavi:
On login, `Auth` sets `secureRouterReferrerPath` to the pathname, but doesn't store any search params or hash that are part of the URL. So `ImplicitCallback` redirects back to the referrer pathname, but it doesn't retain query params or hash.

https://github.com/okta/okta-oidc-js/blob/76e761b15710653970bdd8591e906fce11f38059/packages/okta-react/src/Auth.js#L84
https://github.com/okta/okta-oidc-js/blob/76e761b15710653970bdd8591e906fce11f38059/packages/okta-react/src/ImplicitCallback.js#L43

robertjd:
This change is now available in version 1.0.2 on npm. Thanks again @pallavi !

PR #197, opened 2018-05-02
fix[okta-react]: Retain search and hash on ImplicitCallback redirect

robertjd:
This PR is based on #175 and closes #175 #174

This PR adds a test commit and runs the test in our travis environment.

### DIFF
--- a/packages/okta-react/test/e2e/harness/e2e/page-objects/protected.po.js
+++ b/packages/okta-react/test/e2e/harness/e2e/page-objects/protected.po.js
@@ -14,11 +14,11 @@ import { browser, element, by, ExpectedConditions } from 'protractor';
 
 export class ProtectedPage {
   navigateTo() {
-    return browser.get('/protected');
+    return browser.get('/protected?foo=bar#baz');
   }
 
   waitUntilVisible() {
-    browser.wait(ExpectedConditions.urlContains('/protected'), 5000);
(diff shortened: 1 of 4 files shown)


packages/okta-react/test/e2e/harness/e2e/page-objects/protected.po.js
```diff
@@ -14,11 +14,11 @@ import { browser, element, by, ExpectedConditions } from 'protractor';
 
 export class ProtectedPage {
   navigateTo() {
-    return browser.get('/protected');
+    return browser.get('/protected?foo=bar#baz');
```
jmelberg-okta: If possible, lets try and keep [the same page objects](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-angular/test/e2e/harness/e2e/page-objects/protected.po.ts#L22) throughout our front-end integrations (as we want to use shared specs in the future).

Can this be updated to `/protected/foo?state=bar`? This will allow us to test:
1. Nested paths
2. Retained query params.

robertjd: I'd propose `/protected/foo?state=bar#baz` then because I also want to assert we preserve fragments (we had issues in the past where they were removed, I think it was an authJS issue that got resolved)

Issue #178, opened 2018-04-09
okta-react: ImplicitCallback fails in React.StrictMode because setState is called on unmounted component


mschaeffner:
**How to reproduce:**

* go to react test application and update dependency so that it uses React >= 16.3.0
*  in file `App.js` wrap the `ImplicitCallback` route with a `React.StrictMode` (see https://reactjs.org/docs/strict-mode.html)

change from
```
<Route path='/implicit/callback' component={ImplicitCallback} />
```
to
```
<React.StrictMode>
  <Route path='/implicit/callback' component={ImplicitCallback} />
</React.StrictMode>
```
* Run the application and login with username and password
* The callback is performed and the URL will be http://localhost:3000/implicit/callback
* The application crashes with error: `AuthSdkError: Unable to parse a token from the url`
* The javascript console shows the warning:
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op.

Please check the code for the ImplicitCallback component.
```

ripplegupta03:
I am also facing the same issue, what is the above fix planned to release?
Is there any other workaround for now?

mschaeffner:
The only workaround I know is not using `React.StrictMode` 🙈 
I also like to see this getting merge asap.

robertjd:
This change is now available in version 1.0.2 on npm. Thanks again @mschaeffner !

Issue #138, opened 2018-03-07
okta-react: SecureRoute updates state during an existing state transition


mschaeffner:
This is caused by using `setState` inside `render` and results in following error in the Javascript console:
> Warning: setState(...): Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.

The problem is the call `this.props.auth.login()` in the `render` function of SecureRoute.js.

This error can be reproduced by running the test app in `okta-oidc-js/packages/okta-react/test/e2e/harness/`. 
- start the test app with `npm start`
- go to http://localhost:3000/
- make sure you are not logged in
- click the link _Protected_
- see the error in Javascript console

See also discussion in Okta dev forum:
https://devforum.okta.com/t/react-integration-example/800

robertjd:
This change is now available in version 1.0.2 on npm. Thanks again @mschaeffner !

zg-trinac:
I'm running into this now, seems like the [Okta docs](https://developer.okta.com/code/react/okta_react_sign-in_widget/) is still recommending the same or a similar action for the widget `onAuthRequired`? 

react-router-dom@5.1.2
@okta/okta-react@3.0.2
@okta/okta-signin-widget@4.3.2

swiftone:
@zg-trinac: This exact github issue is a few major versions behind the code you are running, and you're probably seeing a slightly different error. 

Could you open a new issue with details about the error you are seeing?  

When you do so, could you please expand your question a bit, as I'm not clear which recommendation you are asking about.

PR #201, opened 2018-05-04
Fixes for state rendering with strict mode

robertjd:
This is a new PR within this repo (so that IT tests can run on travis).  This PR includes the work from #179 and #139, and modifies the tests to use strict mode so that we don't regress.

This does mean that we are now testing on react 16, with no strategy for testing on 15 for back-compat.  I'm okay with this for now.

Many thanks to @mschaeffner for all the work, sorry this took so long to move forward.

Closes #179 #178 #139 #138

### DIFF
--- a/packages/okta-react/src/SecureRoute.js
+++ b/packages/okta-react/src/SecureRoute.js
@@ -14,6 +14,32 @@ import React, { Component } from 'react';
 import { Route } from 'react-router';
 import withAuth from './withAuth';
 
+
+class RenderWrapper extends Component {
+  checkAuthentication() {
+    if (this.props.authenticated === false) {
+      this.props.login();
+    }
+  }
+
+  componentWillMount() {
+    this.checkAuthentication();
+  }
+
+  componentDidUpdate() {
+    this.checkAuthentication();
+  }
+
+  render() {
+    if (!this.props.authenticated) {
+      return null;
+    }
+
+    const C = this.props.component;
+    return this.props.render ? this.props.render(this.props.renderProps) : <C {...this.props.renderProps} />;
+  }
+}
+
 export default withAuth(class SecureRoute extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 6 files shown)


packages/okta-react/src/SecureRoute.js
```diff
@@ -14,6 +14,32 @@ import React, { Component } from 'react';
 import { Route } from 'react-router';
 import withAuth from './withAuth';
 
+
+class RenderWrapper extends Component {
+  checkAuthentication() {
+    if (this.props.authenticated === false) {
+      this.props.login();
```
jmelberg-okta: Do we need a `fromUri` value here?

packages/okta-react/src/SecureRoute.js
```diff
@@ -14,6 +14,32 @@ import React, { Component } from 'react';
 import { Route } from 'react-router';
 import withAuth from './withAuth';
 
+
+class RenderWrapper extends Component {
```
jmelberg-okta: Any reason you decided to include this class inside of this file? Seems like we could pull this out.

mschaeffner: Because it is only used within this file, and it is strongly related to "SecureRoute". But feel free to pull it out into its own file.

packages/okta-react/src/SecureRoute.js
robertjd: We default to the current location if we don't pass anything:

https://github.com/okta/okta-oidc-js/blob/master/packages/okta-react/src/Auth.js#L84

This actually seems to be the safest approach, because the router does udpate the window location before hitting this wrapped render, and the route path can be a matching expression:

https://reacttraining.com/react-router/web/api/Route/path-string

So by dynamically pulling the window location at this time we're likely to get the actual path that the user should be on after they are sent back 👍

packages/okta-react/src/SecureRoute.js
robertjd: Agreed, it's so coupled it's easier to just see it in the same file

PR #211, opened 2018-05-17
Exclude react-native from CI publish

robertjd:
Will wait on https://github.com/expo/expo/issues/1767 if possible

### DIFF
--- a/scripts/publish_registry.js
+++ b/scripts/publish_registry.js
@@ -26,8 +26,10 @@ dirs.forEach(name => {
   const moduleWithVersion = `${pkg.name}@${pkg.version}`;
 
   if (pkg.name === '@okta/okta-react-native') {
-    console.log(`Skipping ${pkg.name}`)
-    return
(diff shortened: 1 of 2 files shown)


scripts/publish_registry.js
```diff
@@ -25,6 +25,11 @@ dirs.forEach(name => {
   const pkg = require(`${moduleDir}/package`);
   const moduleWithVersion = `${pkg.name}@${pkg.version}`;
 
+  if (pkg.name === '@okta/okta-react-native') {
+    console.log(`Skipping ${pkg.name}`)
```
jmelberg-okta: Lets add a comment here explaining why we need to temporarily skip this package, with a GH issue link or JIRA number for reference.

scripts/publish_registry.js
```diff
@@ -25,6 +25,11 @@ dirs.forEach(name => {
...
+    return
```
jmelberg-okta: Missing `;` at the end of these lines

PR #220, opened 2018-06-07
fix(okta-angular): Fix return type of okta module static method

cmckni3:


robertjd:
Thanks @cmckni3 , is this a fix for Angular 6?  If so is it backcompat with 4?

cmckni3:
It is backwards compatible. It adds the return type so TS compiler does not error when mixing providers.

Issue #215, opened 2018-05-24
After npm run build, Failed minified @okta/jwt-verifier/lib.js


romainlouvet:
I have checked and didn't found this issue

Here is the callstack
`Failed to minify the code from this file:     ./node_modules/@okta/jwt-verifier/lib.js:16
0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/node', '/usr/bin/npm', 'run', 'build' ]
2 info using npm@5.8.0
3 info using node@v6.13.1
4 verbose run-script [ 'prebuild', 'build', 'postbuild' ]
5 info lifecycle Cetautomatix@0.1.0~prebuild: Cetautomatix@0.1.0
6 info lifecycle Cetautomatix@0.1.0~build: Cetautomatix@0.1.0
7 verbose lifecycle Cetautomatix@0.1.0~build: unsafe-perm in lifecycle true
8 verbose lifecycle Cetautomatix@0.1.0~build: PATH: /usr/lib/node_modules/npm/node_modules/npm-lifecycle/node-gyp-bin:/home/elouen/react/toto/ihm-cetautomatix/node_modules/.bin:/home/elouen/.sdkman/candidates/maven/current/bin:/home/elouen/.sdkman/candidates/java/current/bin:/home/elouen/opt/atom:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/local/go/bin:/usr/local/go/bin
9 verbose lifecycle Cetautomatix@0.1.0~build: CWD: /home/elouen/react/toto/ihm-cetautomatix
10 silly lifecycle Cetautomatix@0.1.0~build: Args: [ '-c', 'react-scripts build' ]
11 silly lifecycle Cetautomatix@0.1.0~build: Returned: code: 1  signal: null
12 info lifecycle Cetautomatix@0.1.0~build: Failed to exec build script
13 verbose stack Error: Cetautomatix@0.1.0 build: `react-scripts build`
13 verbose stack Exit status 1
13 verbose stack     at EventEmitter.<anonymous> (/usr/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:283:16)
13 verbose stack     at emitTwo (events.js:106:13)
13 verbose stack     at EventEmitter.emit (events.js:191:7)
[...]

PR #223, opened 2018-06-08
docs[jwt-verifier]: Verifier is only for Node.js, link to Express example

robertjd:
In response to #215

### DIFF
--- a/packages/oidc-middleware/src/oidcUtil.js
+++ b/packages/oidc-middleware/src/oidcUtil.js
@@ -95,7 +95,6 @@ oidcUtil.ensureAuthenticated = (context, options) => {
       ensureLoggedIn(options)(req, res, next);
     } else {
       res.sendStatus(401);
-      next();
     }
   };
 }


packages/jwt-verifier/README.md
```diff
@@ -9,6 +9,10 @@ This library verifies Okta access tokens (issued by Okta authorization servers)
 * Access token is not expired (requires local system time to be in sync with Okta, checks the `exp` claim of the access token).
 * Any custom claim assertions that have been configured.
 
+This library is for Node.js applications and will not compile into a front-end application.  If you need to work with tokens in front-end applications, please see [okta-auth-js](https://github.com/okta/okta-auth-js).
```
jmelberg-okta: nit: To call out this section a bit more, lets add a `>`

> This library is for Node.js applications and will not compile into a front-end application.  If you need to work with tokens in front-end applications, please see [okta-auth-js](https://github.com/okta/okta-auth-js).

Issue #203, opened 2018-05-10
Not compatible with Angular 6


pkspks:
Upgrading to Angular 6 which needs me to upgrade RxJS as well.

npm install output (truncated):
```
npm WARN @okta/okta-angular@1.0.1 requires a peer of @angular/common@^4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @okta/okta-angular@1.0.1 requires a peer of @angular/core@^4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @okta/okta-angular@1.0.1 requires a peer of @angular/platform-browser@^4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @okta/okta-angular@1.0.1 requires a peer of @angular/router@^4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @okta/okta-angular@1.0.1 requires a peer of rxjs@^5.4.3 but none is installed. You must install peer dependencies yourself.
```

App now refuses to start as rxjs 6 have moved package paths.

```
ERROR in node_modules/@okta/okta-angular/dist/okta/services/okta.service.d.ts(4,10): error TS2305: Module '"[LOCAL PATH]/node_modules/rxjs/Observable"' has no exported member 'Observable'.
node_modules/rxjs/Observable.d.ts(1,15): error TS2307: Cannot find module 'rxjs-compat/Observable'.
```

Error goes away after installing `rxjs-compat`

mraible:
Does it work if you install rxjs-compat?

See https://developer.okta.com/blog/2018/05/09/upgrade-to-angular-6 for more info.

pkspks:
@mraible , yes, it works with rxjs-compat. I had updated the issue with this info.

cmckni3:
The RxJS imports need to be fixed in `OktaService`. I will see if I can fix them at some point.

This will require RxJS >= 6.0.0 as a requirement. Not sure how this projects handles versioning in that case.

Maybe mentioning `rjxs-compat` in installation for Angular 6 which requires RxJS 6 is enough for now?

robertjd:
This was fixed in the latest release, 1.0.3

PR #228, opened 2018-06-19
fix(rxjs): Fix RxJS imports

cmckni3:
#203 

- [ ] Update Angular sample application
- [ ] Backwards compatibility strategy


cmckni3:
Need to update the Angular sample application for tests to pass

PR #229, opened 2018-06-19
feat(angular): Add Angular auth interceptor

cmckni3:


mraible:
@cmckni3 It looks like this interceptor auto-configures itself. Is that true?

@robertjd One thing we might want to add is a whitelist/blacklist feature. I don't think we want a Bearer token added to every `HttpClient` request. Third party APIs might fail. For example, when I've added a Bearer token to requests to Giphy, it fails.

cmckni3:
It is opt in via the token `AUTH_API_INTERCEPTOR`.

The issue with interceptors in Angular is that once you opt in, you can’t conveniently skip interceptors. It has be worked around by skipping all interceptors with a different `HttpBackend`

cmckni3:
I can post code later if the skipping part doesn’t make sense.

mraible:
@cmckni3 Right, you can't disable interceptors from a configuration perspective, but you can do something like this:

```ts
import { Injectable } from '@angular/core';
import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
import { Observable } from 'rxjs/Observable';
import 'rxjs/add/observable/fromPromise';
import { OktaAuthService } from '@okta/okta-angular';

@Injectable()
export class AuthInterceptor implements HttpInterceptor {

  constructor(private oktaAuth: OktaAuthService) {
  }

  intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
    return Observable.fromPromise(this.handleAccess(request, next));
  }

  private async handleAccess(request: HttpRequest<any>, next: HttpHandler): Promise<HttpEvent<any>> {
    // Only add to known domains since we don't want to send our tokens to just anyone. 
    // Also, Giphy's API fails when the request includes a token.
    if (request.urlWithParams.indexOf('localhost') > -1) {
      const accessToken = await this.oktaAuth.getAccessToken();
      request = request.clone({
        setHeaders: {
          Authorization: 'Bearer ' + accessToken
        }
      });
    }
    return next.handle(request).toPromise();
  }
}
```

cmckni3:
Yeah did something like that except I used a dependency injected array for the exclusion.

```
    // Skip using Array<RegExp | string>
    const skipInterceptor = this.exclusions.some(url => {
      if (typeof url === 'string') {
        return request.url === url;
      }
      return url.test(request.url);
    });
```

My stackblitz from a few months ago: https://stackblitz.com/edit/angular-cmckni3-http

It's also easy to create a class that inherits from `HttpClient` to skip all interceptors as I mentioned.

```
import { Injectable } from '@angular/core';

import { HttpClient, HttpBackend } from '@angular/common/http';

@Injectable()
export class HttpBackendClient extends HttpClient {
    constructor(handler: HttpBackend) {
        super(handler);
    }
}
```

cmckni3:
For reference: https://github.com/angular/angular/issues/20203#issuecomment-369754776

cmckni3:
Maybe I should close this and keep maintaining my own interceptor?

cmckni3:
Closing due to inactivity

Issue #225, opened 2018-06-15
okta Vuejs implicitCallback components handling error authentication


anasameur:
Problem:

When using the okta-vue package, if a user are authenticated but are not assigned to the client application there is no way to handle this error.

Solution: 
```
export default {
  name: 'ImplicitCallback',
  data: () => ({
    error_message: '',
    error: false,
  }),
  async beforeMount() {
    try {
      await this.$auth.handleAuthentication();
      this.$router.replace({
        path: this.$auth.getFromUri(),
      });
    } catch (e) {
      // handle error
      this.error_message = e.message;
      this.error = true;
    }
  },
  render(createElement) {
    if (this.error) {
      return createElement('div', this.error_message);
    }
    return createElement;
  },
};
```

robertjd:
Hi @anasameur , thank you for this!  There's a similar conversation going on in #234, where @n-zeplo would instead want to redirect the user to the root of the app.  Both are valid use cases, so I'm #236 I'm proposing to show the error by default, but provide a handler that lets that behavior be overridden.  Can you take a look at let me know what you think?

anasameur:
Hi @robertjd Thanks. This was a quick fix to catch the error and display it to the end user as @n-zeplo  said.
I also approve #236 does a better job of handling the error specially when show the error to the end user if this happens, and adding the global error handler onOAuthRedirectError to let developer inject any error callback.

PR #234, opened 2018-06-29
Okta-Vue: Catch error on parsing URI for Access Token 

n-zeplo:
When the URI parsing doesn't work it throws an error and stays on a blank redirect page. This PR will console warn the error but redirect user to the root in order to not stop user experience.

### DIFF
--- a/packages/okta-vue/src/components/ImplicitCallback.vue
+++ b/packages/okta-vue/src/components/ImplicitCallback.vue
@@ -2,10 +2,15 @@
 export default {
   name: 'ImplicitCallback',
   async beforeMount () {
-    await this.$auth.handleAuthentication()
-    this.$router.replace({
-      path: this.$auth.getFromUri()
-    })
+    try {
+      await this.$auth.handleAuthentication()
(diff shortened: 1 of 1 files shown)


robertjd:
Thanks @n-zeplo for the PR!  Landing on a broken blank page is not a great experience, so I can see how redirecting to the root is an improvement.  But I imagine that the user will also be equally confused if they arrive at the root and aren't logged in.  With this PR, how do you plan to give the user a hint that something went wrong during authentication?

robertjd:
Hey again @n-zeplo , it turns out we have a PR at #225 that is also trying to solve this problem.  225 is a route I would take for my own apps, but I don't want to lock anyone into that default behavior.  As such I think it's best to default to showing the error, but still give you control to define your own logic. I've made a PR with this suggestion in #236 , could you take a look and let me know what you think?

n-zeplo:
@robertjd Thanks. This was a quick fix just to catch the error for now so that the app continues to function. I agree that #236 does a better job of handling the error so I will close this PR.

PR #236, opened 2018-07-09
[WIP don't merge] feat[okta-vue]: Add authentication error handler

robertjd:
The default handler will show the error on the page, or the developer can provide their own handler as seen in the example below.

This PR doesn't yet have tests or README changes, this is for discussion on #225 and #234

```javascript
Vue.use(Auth, {
  issuer: sampleConfig.oidc.issuer,
  client_id: sampleConfig.oidc.clientId,
  redirect_uri: sampleConfig.oidc.redirectUri,
  scope: sampleConfig.oidc.scope,
  onAuthenticationError: function (error) {
    console.error(error)
    this.$router.replace('/')
  }
})

```

### DIFF
--- a/packages/okta-vue/src/Auth.js
+++ b/packages/okta-vue/src/Auth.js
@@ -73,6 +73,7 @@ function install (Vue, options) {
       }
     }
   }
+  Vue.prototype.$auth.onAuthenticationError = authConfig.onAuthenticationError
 }
 
 function handleCallback () { return ImplicitCallback }
--- a/packages/okta-vue/src/components/ImplicitCallback.vue
+++ b/packages/okta-vue/src/components/ImplicitCallback.vue
@@ -1,12 +1,31 @@
 <script>
 export default {
   name: 'ImplicitCallback',
-  async beforeMount () {
-    await this.$auth.handleAuthentication()
-    this.$router.replace({
-      path: this.$auth.getFromUri()
-    })
+  data: () => ({
+    error_message: '',
+    error: false,
+  }),
+  async beforeMount() {
+    try {
+      await this.$auth.handleAuthentication();
+      this.$router.replace({
+        path: this.$auth.getFromUri(),
+      });
+    } catch (e) {
+      const handler = this.$auth.onAuthenticationError;
+      if (handler) {
+        return handler.call(this, e);
+      }
+      // Display error message on this callback view
+      this.error_message = e.message;
+      this.error = true;
+    }
   },
-  render () {}
-}
+  render(createElement) {
+    if (this.error) {
+      return createElement('div', this.error_message);
+    }
+    return createElement;
+  },
+};
 </script>


packages/okta-vue/src/Auth.js
```diff
@@ -73,6 +73,7 @@ function install (Vue, options) {
       }
     }
   }
+  Vue.prototype.$auth.onAuthenticationError = authConfig.onAuthenticationError
```
jmelberg-okta: `onAuthenticationError` is a bit generic, as this handler is specifically for the OAuth redirect.

It might make sense to either:
1. Make this a global error handler, so a developer can inject this into any supported error callback
2. Rename this to `onOAuthRedirectError`

packages/okta-vue/src/components/ImplicitCallback.vue
```diff
@@ -1,12 +1,31 @@
...
+        return handler.call(this, e);
+      }
+      // Display error message on this callback view
+      this.error_message = e.message;
+      this.error = true;
+    }
   },
-  render () {}
-}
+  render(createElement) {
```
jmelberg-okta: I'm unsure about the default handler case here, and genuinely feel we should avoid creating elements on the fly. While this is a fallback mechanism for devs who do not implement their own handler, it still puts users in a bad state (non-branded UI and difficult to read error message).

As an alternative, it might make more sense to emit a [custom event](https://vuejs.org/v2/guide/components-custom-events.html) with some redirect logic. In the default case, we could:

1. Listen to the `OAuthRedirectError` event
2. If the event is triggered, navigate to the applications root path `/`  and log the message via console output

packages/okta-vue/src/Auth.js
robertjd: `onOAuthRedirectError` sounds good 👍

packages/okta-vue/src/components/ImplicitCallback.vue
robertjd: IMO it's better to surface the error to the end user if it happens.  A common scenario is this: someone changes the services configuration (e.g. removing a group from the application) and thus accidentally breaks the application for end-users.  If the end user can't see the error, all they can simply report to their support channel is "I'm trying to login and it's not working".  But if they can see the error and report that to their support channel, the resolution can be found more quickly.  

Bump @nbarbettini on this one for your thoughts.  This decision can be applied elsewhere.

jmelberg-okta: I'm not proposing we hide errors from the end user, we just need to specify where we want these errors to output to.

Thinking about this a bit more, IMO these libraries shouldn't dictate how an error message gets to the end user. The responsibility of these libs is to catch any/all errors and expose them in a consumable way.

If we take an event approach, we're able to connect multiple error cases to a single handler, as opposed to telling developers to handle each case independently.

anwarika:
So I've run into the same problem, Will this be fixed? Almost 3 years in the making...

anwarika:
> So I've run into the same problem, Will this be fixed? Almost 3 years in the making...

If anyone else runs into this i just created a custom component handler on `{ path: "/login/callback", component: OktaError }`

and in the OktaError component I handled the errors. 

`<template>
    <div class="position-relative" style="text-align: center; padding-top: 30px">
        {{error_message}}
    </div>

</template>

<script>
    export default {
        name: "OktaError.vue",
        data: () => ({
            error_message: '',
            error: false,
        }),
        async beforeMount() {
            try {
                await this.$auth.handleAuthentication();
                this.$router.replace({
                    path: this.$auth.getFromUri(),
                });
            } catch (e) {
                //TODO Handle error
               this.error_message = e.message
               
            }
        },
    }
</script>`

swiftone:
@anwarika - Thanks for the bump and the workaround.  It looks like fell between the cracks, so thanks to your attention we'll take a look and see about getting this updated and merged.

Note the okta-vue repo has moved to https://github.com/okta/okta-vue, so this work will be moved/copied over to there to be fixed.

Internal ref: OKTA-369846

aarongranick-okta:
Error handling for the callback component is added here: https://github.com/okta/okta-vue/pull/52

To customize error handling, it is recommended to copy the `LoginCallback` component to your own source tree and then modify as needed.

For any problems with `okta-vue`, please file an issue here: https://github.com/okta/okta-vue/issues

PR #237, opened 2018-07-09
SecureRoute accepts exact prop.

### LINKED ISSUE #70: SecureRoute exact=true
In SecureRoute.js https://github.com/okta/okta-oidc-js/blob/8dfb16d3047b0a89a5c90a134285c8ab08de0020/packages/okta-react/src/SecureRoute.js#L57

There is no way to specify additional route props, e.g. `exact`

webdeveloperpr:
Relates to FIX #70 - Seems like this issue has been brought up before, I think it deserves a second look.

The  SecureRoute component is not passing the `exact` prop to the internal Route component making 2 routes rendering on the page which causes infinite re-renders

``` javascript
   <SecureRoute exact path="/" component={withStore(Home)} />
    <Route path="/login" render={() => <Login baseUrl={process.env.OKTA_CLIENT_URL} />} />
```

The exact props tells the router not to render route with path "/" when the url does not exactly match the path, but since we Secure route does not pass the prop, It makes  "/" and "login" routes to render at the same time causing infinite re-renders.

As a User of this library I would expect SecureRoute to accept the same props that the Route does. That is not the case and it will confuse many developers. 

The alternative is to do something like like this:

```javascript
 <Route path="/" exact render={() => {
   return <SecureRoute path="/" component={withStore(
 }} />
 <Route path="/login" render={() => <Login baseUrl={process.env.OKTA_CLIENT_URL} />} />
```

This does work but before you realize that you have to do that you will loose hours or days figuring out why your application is not working as expected.

Also I could use some help with getting this tests passing. Tests that are not related to this PR are failing for some reason.

### DIFF
--- a/packages/okta-react/package.json
+++ b/packages/okta-react/package.json
@@ -75,7 +75,7 @@
       "./test/test-setup.js"
     ],
     "roots": [
-      "<rootDir>/test"
+      "<rootDir>/test/jest"
     ]
   }
 }


robertjd:
Hi @webdeveloperpr , thanks for this! This change certainly makes sense.  Are you trying to run the tests locally?  The tests on this library are mostly e2e tests, and you'll need to have these in your environment first:

```
export ISSUER=https://dev-1234.oktapreview.com
export CLIENT_ID=<from a SPA application>
export USERNAME=<an account that can login to that app>
export PASSWORD=<ditto>
```

I see now that we don't have these test instructions, we'll need to add that.

We did add jest to this library, but we only have one unit test so far.  For this change I'd like to get a test that asserts that we're sending down these props from `SecureRoute` to `Route`.  Any thoughts on a good way to do that?

webdeveloperpr:
So CI will always fail because I'm missing some of the settings so I'm assuming it's ok that those tests are failing.

As for testing components. I have used Enzyme and Jest before. They work really well and could be used to detect the props being passed down.  It would require a bit of configuration and additional packages added to the dev deps. I was hesitant to add them to the PR because I was afraid of doing too much, and not being able to merge.

I could add enzyme to the project and add tests, if this PR is something that you guys would be ok to merge.

packages/okta-react/package.json
```diff
@@ -7,7 +7,8 @@
     "build": "rimraf dist/ && babel src -d dist",
     "build:harness": "npm --prefix test/e2e/harness install",
     "build:package-info": "node ../../util/write-package-info.js . src/packageInfo.js",
-    "jest": "jest test/jest/",
+    "jest": "jest",
+    "jest:watch": "jest --watchAll",
```
webdeveloperpr: Run tests after every test file change

packages/okta-react/package.json
```diff
@@ -61,9 +64,18 @@
     "jest": "^22.4.2",
     "polished": "^1.7.0",
     "protractor": "^5.1.2",
-    "react-dom": "^15.6.1",
+    "react": ">=16",
+    "react-dom": ">=16",
```
webdeveloperpr: add react-dom and react dev-deps as version 16 so tests are compatible with the enzyme adapter. Peer dependencies are still the same.

packages/okta-react/package.json
```diff
@@ -61,9 +64,18 @@
...
     "react-router-dom": "^4.2.2",
     "rimraf": "^2.6.2",
     "styled-components": "^2.1.2"
+  },
+  "jest": {
+    "setupFiles": [
+      "./test/test-setup.js"
```
webdeveloperpr: include the test-setup file in all tests for easier testing. Also sets global enzyme testing functions.

packages/okta-react/test/jest/SecureRoute.spec.js
```diff
@@ -0,0 +1,28 @@
+import { SecureRoute, RenderWrapper } from "../../src/SecureRoute";
+import { MemoryRouter, Route } from "react-router-dom";
+
+describe('SecureRoute', () => {
+  const TestComponent = () => <div>Test Component</div>;
+  const mockProps = {
```
webdeveloperpr: mocks the props that withAuth HOC is passing down to this component.

packages/okta-react/src/SecureRoute.js
```diff
@@ -78,6 +77,15 @@ export default withAuth(class SecureRoute extends Component {
...
+        exact={this.props.exact}
+        path={this.props.path}
+        render={this.createRenderWrapper}
+      />
+    );
   }
-});
+}
+
+export { SecureRoute };
```
webdeveloperpr: this is the unwrapped version of SecureRoute. This makes it easier to test, because now we can mock the props that withAuth was passing before.

webdeveloperpr:
@robertjd Added jest and enzyme and test setup files for component testing. Also added a test for testing the exact prop.

robertjd:
Thanks @webdeveloperpr for the tests, much appreciated 👍 we'll look at this again as soon as possible (it might not be this week, unfortunately).

ericyliu:
what is the status of this? would be a nice to have

amitkapps:
We're planning to use the okta-react component, would love to see this fixed so we can upgrade seamlessly.

jmelberg-okta:
Closing this PR, as it was resolved with #279 and was released under `@okta/okta-react@1.1.0`.

Thanks!

Issue #221, opened 2018-06-08
TypeError: Converting circular structure to JSON


DThanhNg:
I got this error when running a react webapp, trying to login to okta.
I found out that failed in okta-react/dist/Auth.js at 

localStorage.setItem('secureRouterReferrerPath', (0, _stringify2.default)(fromUri ? { pathname: fromUri } : this._history.location));

Console.log() printed that 'fromUri' is a Proxy object.

dhavalsoni2001:
I also get the same error. Any update @DThanhNg ?

DThanhNg:
At the moment, I change in okta-react/dist/Auth.js line 314:
localStorage.setItem('secureRouterReferrerPath', fromUri || this._history.location.pathname);
to:
localStorage.setItem('secureRouterReferrerPath', (0, _stringify2.default)(this._history.location));

I don't know how long it will take Okta team to fix this problem

mraible:
@DThanhNg Send us a PR to fix and we'll try to get it into the next release. ;)

thanaratjaruenying:
Is it fixed? I still get this error when login.

ersinciftci:
Any update on this? I would open a PR that removes `fromUri ? { pathname: fromUri }` part (it works when I test on my local), but I feel like it's there for a reason and needs to be fixed in a better way?

robertjd:
Hi @ersinciftci and @thanaratjaruenying , thanks for confirming this.  Can you show me how you're calling `auth.login(:fromUri)`?  Are you passing it anything as the `fromUri` parameter?  For reference I'm using our [Okta React + Okta Hosted Login Example](https://github.com/okta/samples-js-react/tree/master/okta-hosted-login) and I'm looking at the Navbar component:

https://github.com/okta/samples-js-react/blob/master/okta-hosted-login/src/Navbar.jsx#L36

By default we pass the `/` string for `fromUri`, but I've also tried removing that and passing nothing.  In either case I don't see the error that you are seeing?

Thanks!

ersinciftci:
Hi @robertjd , I think I know what's going on now. In this example https://www.npmjs.com/package/@okta/okta-react#show-login-and-logout-buttons it does `<button onClick={this.props.auth.login}>Login</button>` but it should actually be `<button onClick={() => this.props.auth.login()}>Login</button>`. I changed it in my code and it's fixed.

robertjd:
[...]

PR #238, opened 2018-07-09
docs[okta-react]: Fix login/logout examples in README

robertjd:
Fixes example code that results in errors as described in #221

### DIFF
--- a/packages/okta-react/README.md
+++ b/packages/okta-react/README.md
@@ -100,7 +100,8 @@ export default withAuth(class Home extends Component {
     super(props);
     this.state = { authenticated: null };
     this.checkAuthentication = this.checkAuthentication.bind(this);
-    this.checkAuthentication();
+    this.login = this.login.bind(this);
+    this.logout = this.logout.bind(this);
   }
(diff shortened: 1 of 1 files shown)


packages/okta-react/README.md
```diff
@@ -110,15 +110,27 @@ export default withAuth(class Home extends Component {
     }
   }
 
-  componentDidUpdate() {
+  async login() {
```
jmelberg-okta: We need to make sure these methods are bound in the constructor.

```javascript
constructor(props) {
    super(props);
    this.state = { authenticated: null };
    this.checkAuthentication = this.checkAuthentication.bind(this);
    this.login = this.login.bind(this);
    ...
    this.checkAuthentication();
  }

```

robertjd: 👍 good catch.  I've also removed the last line in the constructor, `this.checkAuthentication();`, as that's not needed (we should be calling that only on the component lifecycle methods, as we do below).  JFYI that I rebased this branch against master and force pushed

Issue #176, opened 2018-04-05
Angular sample crashes when built for prod


nbarbettini:
I tried the latest [Angular custom login](https://github.com/okta/samples-js-angular/tree/master/custom-login) sample:

```
git clone git@github.com:okta/samples-js-angular.git
cd samples-js-angular\custom-login
npm i
ng serve
```

This works fine! However:

```
ng serve --prod
```

Compiles fine, but I get this when I visit `http://localhost:8080`:

![image](https://user-images.githubusercontent.com/7525482/38341721-d5f6e386-382e-11e8-9141-01b745b05198.png)

I suspect something in the configuration is getting mistakenly removed by the tree-shaker, but I don't know enough of what I'm talking about to fix it. 😉

nbarbettini:
Original report: https://devforum.okta.com/t/okta-angular-okta-samples-js-angular-sample-applications/1518

Wrecklaimer:
I'm running into this exact issue after recently updating to the latest version of okta-angular.

At first, I thought it was a dependency conflict with okta-signin-widget. But removing that didn't fix the issue.

Wrecklaimer:
@nbarbettini My issue had something to do with the OktaAuthModule not being configured (undefined). It worked with no issues until a prod build. It seemed pretty obvious once I realized the problem, since "issuer" is part of the Okta config object. I'm guess something similar is happening with the sample app.

GBrawl:
Having the exact same problem here, also tried removing the signin-widget, but apparently that's not where the error occurs. Hoping this gets fixed soon

ijer-web:
Please, solve it!

Frugoletto:
Same exact problem.
It's a blocking issue; please give this a chance to get fixed

Frugoletto:
After some trial and error seems I have managed to find almost a workaround. The problem seems linked to the initialization way of the OktaModule, specifically in the 
```
imports:      [
        ...
        OktaAuthModule.initAuth(environment.oktaConfig),
        ...
    ]
```
section of the AppModule file. 
It seems that the AOT compiler does not fully support provider initialization made in that way, so I tried to comment out the initAuth call and move the code directly in providers section, something like:
```
imports:      [
        ...
        OktaAuthModule,//.initAuth(environment.oktaConfig),
        ...
    ],
    providers: [
      ...
      // okta
      {
        provide: OKTA_CONFIG,
        useValue: environment.oktaConfig
      },
      OktaAuthService,
[...]

PR #243, opened 2018-07-11
fix[okta-angular]: Including Angular 5, 6 as supported

robertjd:
This updates our e2e test to use Angular 6, to assert that the module is compatible with Angular 6

I've also tested Angular 5 in 240c500e5c5c42e0b3cc056b1d76cb7575521595, see the build here: https://travis-ci.org/okta/okta-oidc-js/builds/402857539

Peer dependencies have been updated accordingly.  AFAIK, there a no changes needed to this library, other than the changes for RxJs imports that @cmckni3 provided in 304cd327d7a493e4a93492924f255e8f3d150e7c

Ping @cmckni3 and @mraible to help me sanity check this PR

FYI, the upgrade of the e2e app was achieved with the 4 -> 6 instructions at https://update.angular.io/


cmckni3:
My only complaint is that it claims to support ng v4 but the ng v5 and v6 `@angular/compiler` will not compile for ng v4. A metadata mismatch error will occur. I believe the error occurs for AOT builds which are recommended.

See https://github.com/webcat12345/ngx-ui-switch/issues/219 for an overview of the metadata mismatch error.

robertjd:
Thanks @cmckni3 , help me understand the concern?  Are you saying that AOT builds will fail if `@angular/compiler` is > 4, and is trying to build an AOT app that includes this library?

cmckni3:
Correct ng v4 AOT applications using this library will fail to compile.

My suggestion would be dropping support for ng v4. There may be another solution involving `@angular/compiler` but i’m not sure.

robertjd:
Hey @cmckni3 , so I was sanity checking this just now, and it does seem like this libray will AOT compile with ng4?  You can see a demo here: https://github.com/robertjd/samples-js-angular/commit/a33b8f73fd252b347b1bbca574813c0a9e50d5f4

Do you have a reproduction of the problem that you could show me?

For now I'm inclined to leave the ng4 support, we could always remove it in a new PR.

cmckni3:
Exactly @jmelberg-okta!

cmckni3:
I can come up with some samples tomorrow.

tcoz:
Any timeline for the merge here?

robertdamphousse-okta:
After a few discussions, I've decided that we'll merge these changes for now.  We do have a reproducible AOT failure in #176 and I have a solution that I'll make a second PR for.  We'll also make another PR that adds a full e2e test for versions 4, 5, 6 for future sanity.

cmckni3:
Are you sure 176 is correct? Appears to be an old issue and unrelated.

robertjd:
At the moment it's the only known issue I can reproduce when trying to AOT compile an application with our library, using Angular 4.x and the older version of the compiler.  Do you know of another case?

cmckni3:
No, I thought you were referring to a new issue related to the ng 6 upgrade. #176 has been an issue for awhile. I would recommend the injection token approach or use the environment files.

From what I can see though, the sample should work but the import of the config file may be wrong in the sample.


`import sampleConfig from './.samples.config';` should probably be `
import * as sampleConfig from './.samples.config';`

I would have to dig to find an example of that config file unless you have one. Again, I would recommend the environment file for a static config like this.

robertjd:
Thanks @cmckni3 !  I was also reading about the environment file scheme, we'll be looking into that (we do use it in the tests in this repo).

PR #248, opened 2018-07-20
Run snyk monitor on master commits

robertjd:
See block comment in the script

### DIFF
--- a/scripts/snyk.sh
+++ b/scripts/snyk.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# When packages change their dependency tree, we want to run snyk monitor. This
+# sends the updated dependency tree to snyk for monitoring.  We want to do this
+# when a commit has been made to master, not on PRs because the dependency tree
+# is likely still in flux while review is happening.
+#
+# For simplicity, this script just runs snyk monitor against all packages on
+# master commits, it doesn't try to figure out which packages were updated.
+
+if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+  npm install -g snyk
+  snyk auth $SNYK_API_TOKEN
+  for package in `ls $PWD/packages`;
+  do
+    echo "snyk monitor --org=$SNYK_ORG_ID"
+  done
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ script:
 - npm install
 - npm run bootstrap
 - sh ./scripts/runTests.sh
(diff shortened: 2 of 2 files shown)


scripts/snyk.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# When packages change their dependency tree, we want to run snyk monitor. This
+# sends the updated dependency tree to snyk for monitoring.  We want to do this
+# when a commit has been made to master, not on PRs because the dependency tree
+# is likely still in flux while review is happening.
+#
+# For simplicity, this script just runs snyk monitor against all packages on
+# master commits, it doesn't try to figure out which packages were updated.
+
+if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" ]]; then
```
jmelberg-okta: Travis is throwing an error (and returning `exit 0`) with this line. Can you investigate?

scripts/snyk.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# When packages change their dependency tree, we want to run snyk monitor. This
+# sends the updated dependency tree to snyk for monitoring.  We want to do this
+# when a commit has been made to master, not on PRs because the dependency tree
```
jmelberg-okta: I'm unsure if running **only** on `master` is the best decision here. As we add new packages, we want to make sure we find the dependency vulnerability before it gets merged.

As an alternative, we could have this script listen to `master` and a custom branch `snyk-*`. When we wish to add a new package, our flow will be to validate the dependencies via the `synk-some-description-here` branch, and (if passing), merge to `master`.

robertjd: > I'm unsure if running only on master is the best decision here. As we add new packages, we want to make sure we find the dependency vulnerability before it gets merged.

That concern will be handled separately from this PR.  There is a separate `snyk test` command that can be used for that purpose.  This PR is focused on updating our dep tree with snyk when it does change.

scripts/snyk.sh
robertjd: Show me where that error is? I only see the passing branch test: https://travis-ci.org/okta/okta-oidc-js/builds/406381991

This script technically should not have run yet, since we have not merged this PR to master?

jmelberg-okta: There is an error logged when attempting to evaluate the conditional. See [`L4891`](https://travis-ci.org/okta/okta-oidc-js/builds/406381991#L4891).

robertjd: Ah, thanks!  I just updated this script to be run with bash, looks like it's good now.

PR #250, opened 2018-07-26
Switched from unmaintained passport to passport-next/passport

rwky:
The passport libraries haven't been maintained in a while, this change switches to a fork that is maintained.


rwky:
Tests are failing but they also failed when I ran them on the master branch so I assume there's something wrong with the tests.

PR #247, opened 2018-07-19
feat[okta-react]: Add "security-wrapper" class name to Secure component

robertjd:
Closes #246

PR #256, opened 2018-08-03
Remove wrapping div from Security component's render method

hotgazpacho:
There's no need to throw an additional wrapping div around the children, especially not one without any class names. Just return the children.


hotgazpacho:
Closing in favor of #247

PR #258, opened 2018-08-07
fix: Updates for internal CI scripts

robertjd:
The `setup.sh` script was failing to find globally installed NPM  modules.  Applying the workarounds we've used elsewhere.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -12,9 +12,6 @@ git reset --hard $SHA
 git config --global user.email "oktauploader@okta.com"
 git config --global user.name "oktauploader-okta"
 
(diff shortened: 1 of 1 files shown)


scripts/setup.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export PATH="${PATH}:$(yarn global bin)"
```
jmelberg-okta: Lets make sure we add a comment to be consistent with what we've done in other repos. We should be indicating why this line is required.

```bash
# Add yarn to the $PATH so cli commands do not fail
export PATH="${PATH}:$(yarn global bin)"
```

PR #264, opened 2018-08-16
:seedling: okta-react - Uses configured IdP when requesting token

patrickvinograd:
Passes `idp` parameter specified in constructor through to OktaAuth when requesting token

I realize this could benefit from tests, but:
- The `idp` parameter is well covered in the okta-auth-js test suite
- Given how the tests in this project are structured, I didn't want to impose that everybody who might run it create an IdP in their Okta org.

### DIFF
--- a/packages/okta-react/README.md
+++ b/packages/okta-react/README.md
@@ -192,6 +192,7 @@ Security is the top-most component of okta-react. This is where most of the conf
 - **redirect_uri** (required) - Where the callback handler is hosted
 - **scope** *(optional)*: Reserved or custom claims to be returned in the tokens
 - **response_type** *(optional)*: Desired token grant types
+- **idp** *(optional)*: Custom social or SAML Identity Provider
 - **onAuthRequired** (optional)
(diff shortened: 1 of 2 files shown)


bretterer:
I would like to see a test around this.  I know that it would require and idp to be set for the okta org, so maybe this test can be wrapped in a flag to check for idp first.  I think we should hold off on merging this in until a test is written.

jmelberg-okta:
Hi @patrickvinograd - 

I definitely agree that this should be an available param. To go one step further, there are many available [OAuth 2.0 and OpenID Connect request parameters](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) that our underlying library `okta-auth-js` supports, that are not exposed with `okta-react`.

With #275, you'll be able to pass in the `idp` param via the `login` and `redirect` methods:

```javascript
// Use a custom IdP
await auth.redirect({
  idp: 'idp-unique-identifier'
});
```

Hope the above change will unblock you! If it does not, please let us know.

> Note: I'll keep this PR open until #275 is closed/published.

Issue #261, opened 2018-08-13
Allow additional parameters to be passed in during `/authorize` call


jmelberg-okta:
Similar to the implementation in [Angular](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-angular/src/okta/services/okta.service.ts#L134) and [Vue](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-vue/src/Auth.js#L16), an `additionalParams` object should be an available function argument.

PR #275, opened 2018-09-05
Allow additional params to be passed in for login and redirect methods

jmelberg-okta:
### Description

Allow additional parameters to be passed into the `login` and `redirect` methods. This allows developers to override and/or extend values in their configuration for additional [OAuth 2.0 and OpenID Connect request parameters](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest):
  - `idp`
  - `login_hint`
  - `state`
  - `sessionToken`

#### Usage

```javascript
// Pre-fill username to the primary auth form and redirect back to the root path
await auth.login('/', {
  login_hint: 'my-foo-user'
});

// Similarly
await auth.redirect({
  login_hint: 'my-foo-user'
});
```

### Resolves
- #261

### DIFF
--- a/packages/okta-react/README.md
+++ b/packages/okta-react/README.md
@@ -295,23 +295,29 @@ Retrieves the id token from storage if it exists.
 
 Retrieves the access token from storage if it exists.
 
-#### `auth.login(fromUri)`
+#### `auth.login(fromUri, additionalParams)`
 
-Calls `onAuthRequired` or redirects to Okta if `onAuthRequired` is undefined. This method accepts a `fromUri` parameter to push the user to after successful authentication.
+Calls `onAuthRequired` or redirects to Okta if `onAuthRequired` is undefined. This method accepts a `fromUri` parameter to push the user to after successful authentication, and an optional `additionalParams` object.
+
+For more information on `additionalParams`, see the [`auth.redirect`](#authredirectadditionalparams) method below.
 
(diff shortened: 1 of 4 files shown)


packages/okta-react/README.md
```diff
@@ -295,24 +295,30 @@ Retrieves the id token from storage if it exists.
...
 
 Terminates the user's session in Okta and clears all stored tokens. Accepts an optional `uri` parameter to push the user to after logout.
 
-#### `auth.redirect({sessionToken})`
+#### `auth.redirect(additionalParams)`
+
+Performs a full-page redirect to Okta with optional request parameters.
 
-Performs a redirect to Okta with an optional `sessionToken`.
+The `additionalParams` are mapped to the [AuthJS OpenID Connect Options](https://github.com/okta/okta-auth-js#openid-connect-options). This will override any existing [configuration](#configuration-options). As an example, if you have an Okta `sessionToken`, you can bypass the full-page redirect by passing in this token. This is recommended when using the [Okta Sign-In Widget](https://github.com/okta/okta-signin-widget). Simply pass in a `sessionToken` into the `loginRedirect` method follows:
```
robertjd: We should be linking them here instead, because any of these parameters can be used:

https://developer.okta.com/docs/api/resources/oidc#authorize

We should also update the AuthJS readme to link there as well, rather than duplicating the (now stale) options in that readme.

PR #284, opened 2018-09-10
Expose tokens to middleware

### LINKED ISSUE #226: [oidc-middleware] How to get the access token?
Hey,

I'm trying to implement [this flow](https://developer.okta.com/blog/2018/04/10/oauth-authorization-code-grant-type) in my application and I'm not being able to request an access_token with the code that is returned from OKTA. I imagine that the code was already used(probably by the passport oidc plugin) and because of that, it was expired.

In case this is being returned by the middleware I could not find in the documentation and in the code where it's being stored.

Any help would be appreciated.

jmelberg-okta:
### Description
Attaches the `tokenSet` to the user context object.

#### Resolves
#226

### DIFF
--- a/packages/oidc-middleware/README.md
+++ b/packages/oidc-middleware/README.md
@@ -24,7 +24,7 @@ Need help? Contact [developers@okta.com](mailto:developers@okta.com) or use the 
   - [oidc.ensureAuthenticated({ redirectTo?: '/uri' })](#oidcensureauthenticated-redirectto-uri)
   - [req.isAuthenticated()](#reqisauthenticated)
   - [req.logout()](#reqlogout)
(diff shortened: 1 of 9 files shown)


packages/oidc-middleware/README.md
```diff
@@ -283,10 +295,10 @@ const oidc = new ExpressOIDC({ /* options */ });
 app.use(oidc.router);
 
 function addUserContext(req, res, next) {
-  if (!req.userinfo) return next();
+  if (!req.userContext.userinfo) return next();
```
robertjd: Shouldn't this be `if (!req.userContext)` ?  Also, let's put the return statement on a newline with brackets

Issue #277, opened 2018-09-05
oidc-middleware: library configuration checks


nbarbettini:
To help avoid misconfiguration, this library should implement these checks against common errors: https://oktawiki.atlassian.net/wiki/spaces/PM/pages/552049922/Library+configuration+checks

For this library, these checks apply:

- [x] Issuer
- [x] Client ID
- [x] Client secret

Additionally, make sure these placeholders are always used in documentation or samples:

- [x] `https://{yourOktaDomain}` (not `{yourOktaDomain}.com`)
- [x] `{clientId}`
- [x] `{clientSecret}`

jmelberg-okta:
Resolved with #286

PR #286, opened 2018-09-10
Perform basic configuration checks

jmelberg-okta:
### Description
- Provide basic configuration validation
  - `issuer`
  - `client_id`
  - `client_secret`
  - `redirect_uri`

#### Resolves
- #277

### DIFF
--- a/packages/oidc-middleware/src/ExpressOIDC.js
+++ b/packages/oidc-middleware/src/ExpressOIDC.js
@@ -14,7 +14,12 @@ const EventEmitter = require('events').EventEmitter;
 const _ = require('lodash');
 const oidcUtil = require('./oidcUtil');
 const connectUtil = require('./connectUtil');
-const configUtil = require('./configUtil');
+const {
+  assertIssuer,
+  assertClientId,
+  assertClientSecret,
+  assertRedirectUri
+} = require('./configUtil');
 
 /**
  * Class to easily integrate OpenId Connect with Express
@@ -55,16 +60,16 @@ module.exports = class ExpressOIDC extends EventEmitter {
     } = options;
 
     // Validate the issuer param
-    configUtil.assertIssuer(issuer, options.skipConfigValidation);
+    assertIssuer(issuer, options.skipConfigValidation);
 
     // Validate the client_id param
-    configUtil.assertClientCredentials(client_id);
+    assertClientId(client_id);
 
     // Validate the client_secret param
-    configUtil.assertClientCredentials(client_secret);
+    assertClientSecret(client_secret);
 
     // Validate the redirect_uri param
-    configUtil.assertRedirectUri(redirect_uri);
+    assertRedirectUri(redirect_uri);
 
     // Add defaults to the options
     options = _.merge({
--- a/packages/oidc-middleware/src/ExpressOIDC.js
+++ b/packages/oidc-middleware/src/ExpressOIDC.js
@@ -11,12 +11,10 @@
  */
 
 const EventEmitter = require('events').EventEmitter;
-const uuid = require('uuid');
 const _ = require('lodash');
 const oidcUtil = require('./oidcUtil');
 const connectUtil = require('./connectUtil');
-
-class OIDCMiddlewareError extends Error {}
+const configUtil = require('./configUtil');
 
 /**
  * Class to easily integrate OpenId Connect with Express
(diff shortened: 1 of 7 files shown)


packages/oidc-middleware/src/ExpressOIDC.js
```diff
@@ -36,6 +34,7 @@ module.exports = class ExpressOIDC extends EventEmitter {
    * @param {string} [options.scope=openid] The scopes that will determine the claims on the tokens
    * @param {string} [options.response_type=code] The OpenId Connect response type
    * @param {number} [options.maxClockSkew=120] The maximum discrepancy allowed between server clocks in seconds
+   * @param {Object} [options.skipConfigValidation] Skip configuration validation for specific keys
```
robertjd: I think we should use `testing.disableHttpsCheck` to be consistent with what we've developed for the management SDKs.

packages/oidc-middleware/src/configUtil.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+};
+
+configUtil.assertClientCredentials = (clientIdOrSecret) => {
+  const copyMessage = 'You can copy it from the Okta Developer Console ' +
+    'in the details for the Application you created. ' +
+    `Follow these instructions to find it: ${findAppCredentialsURL}`;
+
+  if (!clientIdOrSecret) {
+    throw new OIDCMiddlewareError('Your client credentials are missing. ' + copyMessage);
```
robertjd: Why are we conflating these two cases into one error message?

packages/oidc-middleware/test/unit/constructor.js
```diff
@@ -14,15 +14,205 @@ describe('new ExpressOIDC()', () => {
...
+  it('should not throw if https issuer validation is skipped', () => {
+    function createInstance() {
+      new ExpressOIDC({
+        issuer: 'http://foo.com',
+        skipConfigValidation: {
+          https: true
+        }
+      });
+    }
+    const errorMsg = 'Your Okta URL must start with https. Current value: http://foo.com. ' +
```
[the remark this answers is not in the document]
jmelberg-okta: I think its better to assert that the error message matches completely, as we'd want tests to fail if we update the error message in our validation checks.

To clean this up, I've concatenated the message to use a shared `const` with the primary error message.

packages/oidc-middleware/src/configUtil.js
```diff
@@ -0,0 +1,66 @@
...
+const findDomainURL = 'https://bit.ly/finding-okta-domain';
+const findAppCredentialsURL = 'https://bit.ly/finding-okta-app-credentials';
+
+configUtil.assertIssuer = (issuer, skip = {}) => {
+  const copyMessage = 'You can copy your domain from the Okta Developer ' +
+    'Console. Follow these instructions to find it: ' + findDomainURL;
+
+  if (!issuer) {
+    throw new OIDCMiddlewareError('Your Okta URL is missing. ' + copyMessage);
+  } else if (!skip.https && !issuer.match(/^https:\/\//g)) {
```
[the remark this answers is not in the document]
jmelberg-okta: Do we? Disabling `https` is an explicit choice for the developer, as they're aware of the implications of doing so. Adding an error message seems like added noise.

packages/oidc-middleware/src/ExpressOIDC.js
jmelberg-okta: I originally thought about this, but ultimately decided against it.

In the future, its likely that we will want to skip `clientId` and `clientSecret` validation, as these may not be known when initializing the `ExpressOIDC` object (Dynamic client registration). Nesting the flags under `testing` would not apply to only testing use cases.

`skipConfigValidation` may not be the best name, but I see it as verbose enough to tell users of this library that they're going against the norm by using this object.

packages/oidc-middleware/src/configUtil.js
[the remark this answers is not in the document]
robertjd: We do still want the warning, to be attention grabbing as possible.

packages/oidc-middleware/src/ExpressOIDC.js
robertjd: We had a long conversation about this, and in the end we decided this should go under the `testing` key to be very clear that this is for a testing case and should be avoided for any kind of real production situation.  Hence the warning I mention above.  We could still add some other key in the future to address the client ID concerns you raise, but we'd still want to break out this https disabling to the `testing` key.

PR #296, opened 2018-09-17
Minor cleanup in runIntegrationTests.sh

bdemers:
Updating TCK_VERSION to include `-SNAPSHOT` as this version could be something like `1.2.3` or `4.5.6-SNAPSHOT`


jmelberg-okta:
Changes captured with #326

PR #302, opened 2018-09-18
fix[okta-react]: Fixes login_required error

manueltanzi-okta:
### Description

### Reviewers

@robertjd @jmelberg-okta @nbarbettini

### DIFF
--- a/packages/okta-react/src/Auth.js
+++ b/packages/okta-react/src/Auth.js
@@ -71,13 +71,27 @@ export default class Auth {
   }
 
   async getIdToken() {
-    const idToken = await this._oktaAuth.tokenManager.get('idToken');
-    return idToken ? idToken.idToken : undefined;
+    try {
+      const idToken = await this._oktaAuth.tokenManager.get('idToken');
+      return idToken.idToken;
+    } catch (err) {
+      // The user no longer has an existing SSO session in the browser.
+      // (OIDC error `login_required`)
+      // Ask the user to authenticate again.
+      return undefined;
+    }
   }
 
   async getAccessToken() {
-    const accessToken = await this._oktaAuth.tokenManager.get('accessToken');
-    return accessToken ? accessToken.accessToken : undefined;
+    try {
+      const accessToken = await this._oktaAuth.tokenManager.get('accessToken');
+      return accessToken.accessToken;
+    } catch (err) {
+      // The user no longer has an existing SSO session in the browser.
+      // (OIDC error `login_required`)
+      // Ask the user to authenticate again.
+      return undefined;
+    }
   }
 
   async login(fromUri, additionalParams) {
--- a/packages/okta-react/test/jest/auth.test.js
+++ b/packages/okta-react/test/jest/auth.test.js
@@ -38,11 +38,24 @@ const mockAuthJsInstance = {
   }
 };
 
-AuthJS.mockImplementation(() => {
-  return mockAuthJsInstance
-});
+const mockAuthJsInstanceWithError = {
+  userAgent: 'okta-auth-js',
+  tokenManager: {
+    get: jest.fn().mockImplementation(() => {
+      throw new Error();
+    })
+  },
+  token: {
+    getWithRedirect: jest.fn()
+  }
+};
 
 describe('Auth component', () => {
+  beforeEach(() => {
+    AuthJS.mockImplementation(() => {
+      return mockAuthJsInstance
+    });
+  });
   test('sets the right user agent on AuthJS', () => {
     const auth = new Auth({
       issuer: 'https://foo/oauth2/default'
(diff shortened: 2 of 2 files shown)


packages/okta-react/test/jest/auth.test.js
```diff
@@ -122,4 +135,27 @@ describe('Auth component', () => {
       foo: 'bar'
     });
   });
+  test('returns isAuthenticated true when TokenManager can get the accessToken', async () => {
```
robertjd: Suggested wording, `'isAuthenticated() returns true when the TokenManager returns an access token'`

packages/okta-react/test/jest/auth.test.js
```diff
@@ -122,4 +135,27 @@ describe('Auth component', () => {
...
+  });
+});
+
+describe('Auth component', () => {
+  beforeEach(() => {
+    AuthJS.mockImplementation(() => {
+      return mockAuthJsInstanceWithError
+    });
+  });
+  test('returns isAuthenticated false when tokenManager throws an error', async () => {
```
robertjd: Similar suggestion as above.. `isAuthenticated() returns false when..`

packages/okta-react/src/Auth.js
```diff
@@ -53,7 +53,11 @@ export default class Auth {
   async isAuthenticated() {
     // If there could be tokens in the url
     if (location && location.hash && containsAuthTokens.test(location.hash)) return null;
-    return !!(await this.getAccessToken()) || !!(await this.getIdToken());
+    try {
+      return !!(await this.getAccessToken()) || !!(await this.getIdToken());
```
jmelberg-okta: Curious why we're trying to catch errors here, but neglecting the root cause of the error at the [source: `getAccessToken` and `getIdToken` methods](https://github.com/okta/okta-oidc-js/blob/e0d60e45a75d35936b1bf92af4cbe65c87cba648/packages/okta-react/src/Auth.js#L82).

Since the error is thrown during AuthJS's [`refresh`](https://github.com/okta/okta-auth-js/blob/master/lib/TokenManager.js#L127-L135) call due to the `prompt=none` param, we should probably add the `try catch` logic to the getter methods.

Then, users of the `getAccessToken` and `getIdToken` methods don't need to wrap it in a `try catch`, as the contract is:

- If you receive a token, it is valid
- If you do not receive a token, your session has ended.

manueltanzi-okta: The idea was to handle the exception for the `isAuthenticated` function because the contract of that function is that returns true when you are authenticated and false otherwise, and it is used for routing. 
While for the other functions, let the error be visible to the developer, so they know why the `getAccessToken` is not giving a token even if it is available in the localStorage.

However, looking at the contract you posted above, makes sense what you are saying, because anyway the token in localStorage would not be valid when this exception is thrown. Will update!

rodoabad:
Sweet, now that this has been merged deployed to master how long is your ETA for a release on NPM?

packages/okta-react/src/Auth.js
```diff
@@ -71,13 +71,27 @@ export default class Auth {
...
-    const idToken = await this._oktaAuth.tokenManager.get('idToken');
-    return idToken ? idToken.idToken : undefined;
+    try {
+      const idToken = await this._oktaAuth.tokenManager.get('idToken');
+      return idToken.idToken;
+    } catch (err) {
+      // The user no longer has an existing SSO session in the browser.
+      // (OIDC error `login_required`)
+      // Ask the user to authenticate again.
+      return undefined;
```
rodoabad: Just wondering, is this the only error that will be returned if promise is rejected?

The implementation pretty much says if a promise is rejected, regardless of what it is, ask the user to authenticate again (by returning `undefined).

rodoabad: Just wondering, is this the only error that will be returned if promise is rejected?

The implementation pretty much says if a promise is rejected, regardless of what it is, ask the user to authenticate again (by returning `undefined).

Wouldn't it be better to *catch* errors that you know and then reject the rest that you don't know how to *catch* yet?

```js
try { }
catch(error) {
  if(error.type === 'login_required') {
    return handleLoginRequiredError(error); // returns `undefined`
  }

  return Promise.reject(error);
}
```

Or you could use an `errorHandlerMap` if there's more than one error that you know.

robertjd: Hi @rodoabad , good question.  You're right that the current approach is opinionated.  If this ends up being too limiting, please let us know and we'll make the error more workable.

evbots: @robertjd @rodoabad @jmelberg-okta @manueltanzi-okta 

How do I log the user out after a session has expired? 

In my Okta org I set tokens to expire after 5 minutes, and set sessions to expire after 6 minutes. When I am testing, I attempt to access a `<SecureRoute>` after the session has expired. The secure route is loaded (even though the session has expired), and inside the component `auth.isAuthenticated()` is called, which internally catches the `login_required` error and returns `undefined`. At this point, if I call `auth.logout()`, I get an error:
```
Unhandled Rejection (AuthApiError): Not found: Resource not found: me (Session)
```

So, should I just redirect to my `/login` path even though the auth module does not consider me to be logged out? What's the way I should be approaching this? I just want my application state to reflect the fact that my Okta session has ended.

PR #304, opened 2018-09-19
fix[okta-vue]: Fixes login_required error

manueltanzi-okta:
### Description

Also makes the isAuthenticated function to use the getAccessToken() and getIdToken() functions instead of using directly the tokenManager, like we do for the other SDKs.

### Reviewers

@robertjd @jmelberg-okta @nbarbettini 

### Additional Reviewers

@sirgalleto

### DIFF
--- a/packages/okta-vue/src/Auth.js
+++ b/packages/okta-vue/src/Auth.js
@@ -27,7 +27,7 @@ function install (Vue, options) {
       await oktaAuth.signOut()
     },
     async isAuthenticated () {
-      return !!(await oktaAuth.tokenManager.get('accessToken')) || !!(await oktaAuth.tokenManager.get('idToken'))
+      return !!(await this.getAccessToken()) || !!(await this.getIdToken())
     },
     async handleAuthentication () {
       const tokens = await oktaAuth.token.parseFromUrl()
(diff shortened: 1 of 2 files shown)


packages/okta-vue/src/Auth.js
```diff
@@ -27,7 +27,7 @@ function install (Vue, options) {
       await oktaAuth.signOut()
     },
     async isAuthenticated () {
-      return !!(await oktaAuth.tokenManager.get('accessToken')) || !!(await oktaAuth.tokenManager.get('idToken'))
+      return !!(await this.getAccessToken()) || !!(await this.getIdToken())
```
sirgalleto: So, I'm wondering how this condition will check if the token is expired

jmelberg-okta: Great question @sirgalleto -

Token expiration is handled automatically by our dependency `okta-auth-js` via [automatic token renewal](https://github.com/okta/okta-auth-js/#the-tokenmanager).

If a token is returned by `okta-vue`'s `getAccessToken()` or `getIdToken()` methods, we're stating it hasn't expired and has been [validated](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation) (for identity tokens).

jmelberg-okta: Exactly.

When `autoRenew` is disabled, we're not going to try and renew the token, and that extra API call to renew it will not happen (and the error will not be thrown). Instead, we [remove the token](https://github.com/okta/okta-auth-js/blob/master/lib/TokenManager.js#L111) from storage, resulting in an `undefined` return value.

packages/okta-vue/src/Auth.js
```diff
@@ -42,11 +42,21 @@ function install (Vue, options) {
       return path
     },
     async getIdToken () {
-      const idToken = await oktaAuth.tokenManager.get('idToken')
```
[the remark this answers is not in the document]
robertjd: I do think this reads better (using `const` inside `try`, return inside of try) 👍 .  Regarding comments, we do want to give some information on what's happening, similar to what we're doing in #307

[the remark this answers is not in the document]
manueltanzi-okta: Will update, @sirgalleto thank you for contributing!

packages/okta-vue/src/Auth.js
sirgalleto: Nice, thank you for that context. Seems like you have a great team.

packages/okta-vue/src/Auth.js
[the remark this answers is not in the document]
sirgalleto: And something that is in my mind right now is the use of `idToken. idToken`. Probably we can rename the returned object of `tokenManager.get('idToken')` but now I'm not sure what's the name of that abstraction. 

Is the returned object like a token object with only the prop you send to the get method?

sirgalleto:
Is this already available in the npm registry?

manueltanzi-okta:
@sirgalleto Yepp! @okta/okta-vue@1.0.4

Issue #297, opened 2018-09-17
Cleanup packages/tests


jmelberg-okta:
### Description

This repository was designed as a monorepo using [`lerna`](https://github.com/lerna/lerna), yet a number of `lerna` toolkit items are not utilized. To ease developer contribution, this repository should allow developers to:

1. Easily run and create unit, integration, and end-to-end tests
    - Running tests locally require _custom_ setup for each package. Using a shared configuration would simplify the overhead required.
2. Shared dependencies should be pulled to the root
    - `lerna` pulls shared `devDependencies` into the root of the project, which has a [number of benefits](https://github.com/lerna/lerna#common-devdependencies). Currently, each package contains its own unique set of dependencies, oftentimes using stale/different versions.

jmelberg-okta:
Resolved with #314

PR #306, opened 2018-09-19
Build constancy improvements

bdemers:
* CI now uses npm 6.2.0
* dotenv is uses to help run our ITs (same way the e2e tests work in the TCK)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ packageInfo.js
 packageInfo.ts
 coverage
 *.tgz
+.idea/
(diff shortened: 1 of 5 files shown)


jmelberg-okta:
Thanks @bdemers!

In theory, there should be one shared configuration file across all our packages, which lives at the root of this repo. This of course requires a bit of refactoring, and we're tracking test-related work in  #297.

RE: `npm` version upgrade - We're unsure if sticking with `npm` is the correct approach. We're exploring moving off `npm` in favor of `yarn`.

That being said, lets hold off on this PR until some decisions on our side have been made.

jmelberg-okta:
Closing with respect to #314

PR #307, opened 2018-09-20
fix[okta-angular]: Fixes login_required error

manueltanzi-okta:
### Description

### Reviewers

@robertjd  @jmelberg-okta @nbarbettini

### DIFF
--- a/packages/okta-angular/src/okta/services/okta.service.ts
+++ b/packages/okta-angular/src/okta/services/okta.service.ts
@@ -88,16 +88,30 @@ export class OktaAuthService {
      * Returns the current accessToken in the tokenManager.
      */
     async getAccessToken(): Promise<string | undefined>  {
-      const accessToken = await this.oktaAuth.tokenManager.get('accessToken');
-      return accessToken ? accessToken.accessToken : undefined;
+      try {
+        const accessToken = await this.oktaAuth.tokenManager.get('accessToken');
+        return accessToken.accessToken;
+      } catch (err) {
+        // The user no longer has an existing SSO session in the browser.
+        // (OIDC error `login_required`)
+        // Ask the user to authenticate again.
+        return undefined;
+      }
     }
 
     /**
      * Returns the current idToken in the tokenManager.
      */
     async getIdToken(): Promise<string | undefined> {
-      const idToken = await this.oktaAuth.tokenManager.get('idToken');
-      return idToken ? idToken.idToken : undefined;
+      try {
+        const idToken = await this.oktaAuth.tokenManager.get('idToken');
+        return idToken.idToken;
+      } catch (err) {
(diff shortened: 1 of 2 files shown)


packages/okta-angular/src/okta/services/okta.service.ts
```diff
@@ -88,16 +88,30 @@ export class OktaAuthService {
      * Returns the current accessToken in the tokenManager.
      */
     async getAccessToken(): Promise<string | undefined>  {
-      const accessToken = await this.oktaAuth.tokenManager.get('accessToken');
-      return accessToken ? accessToken.accessToken : undefined;
+      try {
```
cmckni3: sweet. I would have rewritten this to use an `Observable` to make it more cohesive in Angular applications.

```
import { from as observableFrom, Observable, of as observableOf } from 'rxjs';
import { catchError, map } from 'rxjs/operators';

getAccessToken(): Observable<string | undefined> {
  return observableFrom(this.oktaAuth.tokenManager.get('idToken')).pipe(
    map(accessToken => accessToken.accessToken),
    catchError(_err => observableOf(undefined))
  );
}
```

packages/okta-angular/src/okta/services/okta.service.ts
```diff
@@ -88,16 +88,30 @@ export class OktaAuthService {
...
+      }
     }
 
     /**
      * Returns the current idToken in the tokenManager.
      */
     async getIdToken(): Promise<string | undefined> {
-      const idToken = await this.oktaAuth.tokenManager.get('idToken');
-      return idToken ? idToken.idToken : undefined;
+      try {
```
cmckni3: See comment about `getAccessToken` using`Observable`s

packages/okta-angular/src/okta/services/okta.service.ts
robertjd: Thanks @cmckni3.  My understanding with promise vs. observable is that observables should be used when more than one value will be returned over time?   I'm not sure that this getter method falls into that category?  Once we return a token we don't intend to do any more work (or return any more tokens) until the method is called again.

cmckni3: On the values over time: yes and no. Observables shine in the async world because of the push pull and cancelable features. Observables can emit only one value (examples: when `from` is used to convert from a Promise or from a value that is not an array, an explicit `take`/`takeUntil`/`takeWhile` operator).

My main concern with this is mixing `async`/`await` with Observables is problematic (from the consumer application side). In this case it's fine since it is catching all errors but maybe the error should actually be returned. Different errors for different situations.

Implementing this Observable pattern opens up the door to a polling of the token. For example, running a HTTP request every few seconds to see if the token has expired. Can expose this method to the outside world too.

Just my thoughts and experience from using RxJS in applications.

PR #309, opened 2018-09-20
Adds support to run the https mode of the TCK

bdemers:
Moved common TCK logic to scripts/tck.sh
The TCK uses a self signed cert, which is unpacked to `tck-keystore.pem` and added to the env via NODE_EXTRA_CA_CERTS

**NOTE:** CI will fail until okta/okta-oidc-tck#40 is merged


jmelberg-okta:
Included change with #326

PR #313, opened 2018-09-22
release[okta-angular][okta-react][okta-vue]: Publish new versions

manueltanzi-okta:
### Description

- Publish okta-angular v1.0.4
- Publish okta-react v1.1.1
- Publish okta-vue v1.0.4

### Reviewers

@robertdamphousse-okta @jmelberg-okta @nbarbettini

### DIFF
--- a/packages/okta-vue/CHANGELOG.md
+++ b/packages/okta-vue/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug fixes
 
-- [`95c3e62`](https://github.com/okta/okta-oidc-js/commit/dbfb7de3b41e932559ffa70790eeeca1dd30c270) - Fixes login_required error.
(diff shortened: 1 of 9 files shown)


packages/okta-vue/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 1.0.4
+
+### Bug fixes
+
+- [`95c3e62`](https://github.com/okta/okta-oidc-js/commit/dbfb7de3b41e932559ffa70790eeeca1dd30c270) - Fixes login_required error.
```
jmelberg-okta: Lets be a little more verbose here - as the fix satisfied a bigger problem. What do you think about:

> Fixes an issue where the library would enter an error state when attempting to renew expired tokens

manueltanzi-okta: Makes sense but I would leave the name of the error (`login_required`), since many people can identify their problem easily thanks to that ^^

PR #377, opened 2019-01-24
chore(middleware): Adds appBaseUrl, removes redirect_uri

swiftone:
BREAKING CHANGE:

Configurations now require appBaseUrl.
redirect_uri is deprecated

## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [X] Feature
- [X] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [X] Documentation changes

The ExpressOIDC() constructor expects to be passed a `redirect_uri` param

The optional `routes` option has a `callback` param that can have a `defaultRedirect` option
Issue Number: N/A

## What is the new behavior?
The ExpressOIDC() constructor expects an `appBaseUrl` param (it will construct default `redirect_uri` from that and the `routes`, be those default or custom)
* `routes.callback` is now `routes.loginCallback`
* `routes.callback.defaultRedirect` is now `routes.loginCallback.afterCallback`

- [X] Yes
- [ ] No

See "Updating" in the README

## Other information

There will be a follow-up breaking change to add a new default `/logout` route

## Reviewers
@manueltanzi-okta @robertdamphousse-okta

### DIFF
--- a/packages/oidc-middleware/README.md
+++ b/packages/oidc-middleware/README.md
@@ -11,6 +11,7 @@
 * [Getting started](#getting-started)
 * [Usage guide](#usage-guide)
 * [API reference](#api-reference)
+* [Upgrading](#upgrading)
 * [Contributing](#contributing)
 
 This package makes it easy to get your users logged in with Okta using OpenId Connect (OIDC).  It enables your Express application to participate in the [authorization code flow][auth-code-docs] flow by redirecting the user to Okta for authentication and handling the callback from Okta.  Once this flow is complete, a local session is created and the user context is saved for the duration of the session.
@@ -19,11 +20,12 @@ This package makes it easy to get your users logged in with Okta using OpenId Co
 
 This library uses semantic versioning and follows Okta's [library version policy](https://developer.okta.com/code/library-versions/).
 
-:heavy_check_mark: The current stable major version series is: 1.x
+:heavy_check_mark: The current stable major version series is: 2.x
 
 | Version | Status                    |
 | ------- | ------------------------- |
-| 1.x     | :heavy_check_mark: Stable |
+| 2.x     | :heavy_check_mark: Stable |
+| 1.x     | :x: Deprecated            |
 | 0.x     | :x: Retired               |
 
 The latest release can always be found on the [releases page][github-releases].
@@ -37,7 +39,9 @@ If you run into problems using the SDK, you can:
 
 ## Getting started
 
-Installing the Okta Node JS OIDC MIddlware in your project is simple.
+See [Upgrading](#upgrading) for information on updating to the latest version of the library.
+
+Installing the Okta Node JS OIDC Middlware in your project is simple.
 
 ```sh
 # npm
@@ -67,7 +71,7 @@ const oidc = new ExpressOIDC({
   issuer: 'https://{yourOktaDomain}/oauth2/default',
   client_id: '{clientId}',
   client_secret: '{clientSecret}',
-  redirect_uri: 'http://localhost:3000/authorization-code/callback',
+  appBaseUrl: '{appBaseUrl}',
   scope: 'openid profile'
 });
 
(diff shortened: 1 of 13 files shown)


packages/oidc-middleware/src/ExpressOIDC.js
```diff
@@ -68,28 +71,35 @@ module.exports = class ExpressOIDC extends EventEmitter {
...
+    // Build redirect uri unless explicitly set
+    options.loginRedirectUri = loginRedirectUri || `${appBaseUrl}${options.routes.loginCallback.path}`;
+
+    // Validate the redirect_uri param
+    assertRedirectUri(options.loginRedirectUri);
+
     const context = {
-      options
+      options,
+      emitter: this
```
manueltanzi-okta: Curious, what's the reason for this change? To make it more explicit?

yarn.lock
```diff
@@ -566,6 +566,11 @@
     pouchdb-collections "^1.0.1"
     tiny-queue "^0.2.1"
 
+"@okta/configuration-validation@^0.1.1":
+  version "0.1.1"
```
manueltanzi-okta: Wonder, should this be 0.2.0?

packages/oidc-middleware/README.md
```diff
@@ -139,10 +143,11 @@ Required config:
 * **issuer** - The OIDC provider (e.g. `https://{yourOktaDomain}/oauth2/default`)
 * **client_id** - An id provided when you create an OIDC app in your Okta Org
 * **client_secret** - A secret provided when you create an OIDC app in your Okta Org
-* **redirect_uri** - The callback for your app. Locally, this is usually `http://localhost:3000/authorization-code/callback`. When deployed, this should be `https://{yourProductionDomain}/authorization-code/callback`.
+* **appBaseUrl** - The base scheme, host, and port (if not 80/443) of your app, not including any path (e.g. http://localhost:3000, not http://localhost:3000/ )  You may specific `loginRedirectUri` to override this setting if you redirect to other apps.
```
manueltanzi-okta: I think `specify` instead of `specific`.
nit: Also, not sure if we should say `to override this setting`, since they should have different roles. Maybe we can instead explain its standalone behavior, since the override is already explained in `loginRedirectUri`.

packages/oidc-middleware/README.md
```diff
@@ -139,10 +143,11 @@ Required config:
 * **issuer** - The OIDC provider (e.g. `https://{yourOktaDomain}/oauth2/default`)
 * **client_id** - An id provided when you create an OIDC app in your Okta Org
```
manueltanzi-okta: Let's leave this for the future, but probably at some point we should standardize to all camelCase or all underscore_case.

packages/oidc-middleware/README.md
```diff
@@ -366,6 +376,20 @@ Once you have done that you can read the documentation on the [request][] librar
 [openid-client]: https://github.com/panva/node-openid-client
 [request]: https://github.com/request/request
 
+### Upgrading 
+
+#### from 1.x to 2.x
```
manueltanzi-okta: Should we mention they need to add `appBaseUrl`?

yarn.lock
swiftone: It should...and I have no idea why it's not updating.

packages/oidc-middleware/src/ExpressOIDC.js
swiftone: Currently, only the ExpressOIDC object or something with a reference to it can emit.  

Now that we can have errors on more than setup, (such as an error during token revoke) we need a way to emit errors on those, and those don't happen in ExpressOIDC, but rather in another call that ExpressOIDC calls.  

We could have THOSE be emitters and have ExpressOIDC listen and re-emit, or we can just pass the emitter down so those can call the emitter directly when needed.

packages/oidc-middleware/README.md
swiftone: Agreed

packages/oidc-middleware/README.md
swiftone: I think this is a transition in process

packages/oidc-middleware/README.md
swiftone: That's the first line under the next subheading

swiftone: ...or I'm looking at the wrong version of the file.  I'll update to add adding appBaseUrl and removing redirect_uri

PR #419, opened 2019-04-03
Remove peer dependency restriction to allow latest version of Angular (7)

aarongranick-okta:
## PR Checklist

- [X ] The commit message follows our [guidelines](/okta/okta-oidc-js/blob/master/CONTRIBUTING.md#commit)
- [ N/A ] Tests for the changes have been added (for bug fixes / features)
- [ N/A ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [ X ] Other... Please describe:
Relax peer dependency for angular 7+

A peer dependency warning is shown if using angular >= 7

Issue Number: N/A

## What is the new behavior?
No peer warning is shown for higher version of Angular

- [ ] Yes
- [X ] No

### DIFF
--- a/packages/okta-angular/package.json
+++ b/packages/okta-angular/package.json
@@ -48,10 +48,10 @@
     "typescript": "^2.5.2"
   },
   "peerDependencies": {
-    "@angular/common": ">=4 <7",
-    "@angular/core": ">=4 <7",
-    "@angular/platform-browser": ">=4 <7",
-    "@angular/router": ">=4 <7",
-    "rxjs": ">=5.4.3 <7"
+    "@angular/common": ">=4",
+    "@angular/core": ">=4",
+    "@angular/platform-browser": ">=4",
+    "@angular/router": ">=4",
+    "rxjs": ">=5.4.3"
   }
 }


packages/okta-angular/package.json
```diff
@@ -48,10 +48,10 @@
...
-    "@angular/common": ">=4 <7",
-    "@angular/core": ">=4 <7",
-    "@angular/platform-browser": ">=4 <7",
-    "@angular/router": ">=4 <7",
-    "rxjs": ">=5.4.3 <7"
+    "@angular/common": ">=4",
+    "@angular/core": ">=4",
+    "@angular/platform-browser": ">=4",
+    "@angular/router": ">=4",
+    "rxjs": ">=5.4.3"
```
swiftone: We've clearly fallen behind on staying on top of this in the past, but do we want to make it open ended like this?

aarongranick-okta: I think its OK to take an "optimistic" approach here. When/if there appears some future incompatibility we can either fix or restrict.

swiftone: But if we fix-restrict after the fact, users won't get the info that there's a problem unless they upgrade.

What's the case in favor of the optimistic approach?

aarongranick-okta: Basically, we have tested and are aware of no incompatibility with any current versions. We could say " < 8" but there is no Angular 8. I assume there is a reason we are not compatible with Angular 3 which is why we specify >= 4. Otherwise the peer dependency emits a warning to ensure the developers have added a dependency on Angular themselves (because we don't include it).

robertjd: @aarongranick-okta and I chatted about this.  The decision is to remove the upper bound so that we aren't pre-emptively causing peer warnings, with a follow up task (OKTA-220767) to create a cron job in travis that will run our tests against the beta channel of the next major angular version, so that we can detect problems before the next major release happens.  Then we can decide to invest time in a fix or place an upper bound on the range until we have time to fix.

PR #436, opened 2019-04-24
Added support for an includes operator for assertClaims (which could be arrays or space separated strings)

dogeared:
## PR Checklist

**NOTE:** I wanted to get eyes on this before updating the docs. If it's approved conceptually, I will update the docs.

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Currently, only simple claim/value pairs can be used in the `assertClaims` options object

## What is the new behavior?

Now, you can provide additional hints in `assertClaims` like:

```
const verifier = new OktaJwtVerifier({
      issuer: ISSUER,
      clientId: CLIENT_ID,
      assertClaims: {
        'groups.includes': ['Everyone', 'Another']
      }
    });
```
This will ensure that the `groups` claim (an array) includes the values `Everyone` and `Another`

This would also work for a claim with a space separated value like:

```
const verifier = new OktaJwtVerifier({
      issuer: ISSUER,
      clientId: CLIENT_ID,
      assertClaims: {
        'scp.includes': ['promos:write', 'promos:delete']
      }
    });
```

This ensures that the `scp` claim (a space separated string) includes the values `promos:write` and `promos:delete`

- [ ] Yes
- [x] No


packages/jwt-verifier/lib.js
```diff
@@ -51,10 +51,28 @@ class OktaJwtVerifier {
         delete jwt.body;
         const errors = [];
         for (let claim of Object.keys(this.claimsToAssert)) {
-          const actualValue = jwt.claims[claim];
           const expectedValue = this.claimsToAssert[claim];
-          if (actualValue !== expectedValue) {
+          let op = undefined;
```
swiftone: Not to unduly criticize unrequested bounties of free code upgrades, but I'd love to see this new code pulled out into some smaller separate functions.

packages/jwt-verifier/test/spec/verifier.spec.js
```diff
@@ -147,6 +147,63 @@ describe('Jwt Verifier', () => {
       )));
   });
 
+  it('should only allow includes operator for custom claims', () => {
```
swiftone: Why no test of the happy path?

Also (untested) - are these tests going to handle the async well?  You're not using done() nor declaring the test function to be async, so I'd think the actual expects() would potentially run AFTER the test runner was complete.

dogeared: I was just mimicking the existing test for assertClaims. Maybe they all need a reworking?

dogeared: I experimented with testing the happy path, but it requires actually getting information into a real access token.

I'd rather write some unit tests with fixtures rather than relying on the presence of a real okta org. I'll get on that.

dogeared:
@swiftone - Take a look at the tests again when you have a chance. I separated it out into those tests that require api calls and a set of tests for `assertClaims` that use a stubbed out verifier and don't make any calls over the wire.

dogeared:
@swiftone bump

dogeared:
@swiftone - I extracted the code. Let me know what you think. If it's heading in the right direction, I'll update docs.

swiftone:
LGTM, thanks for the changes.  Go ahead with the docs and poke me when it's ready for approval.  Much appreciated, we have more than one issue this will impact.

dogeared:
I updated the README in the `jwt-verifier` package. Is there anywhere else this needs to be documented? If not, It's GTG!

packages/jwt-verifier/lib.js
```diff
@@ -18,6 +18,47 @@ const {
   assertClientId
 } = require('@okta/configuration-validation');
 
+class AssertedClaimsVerifier {
+  constructor() {
+    this.errors = [];
+  }
+
+  extractOperator(claim) {
+    let idx = claim.indexOf('.');
```
mraible: const instead of let

packages/jwt-verifier/lib.js
```diff
@@ -18,6 +18,47 @@ const {
...
+    if (idx >= 0) {
+      return claim.substring(idx + 1);
+    }
+    return undefined;
+  }
+
+  extractClaim(claim) {
+    let idx = claim.indexOf('.');
```
mraible: const is the new hotness

dogeared: fixed both

dogeared:
bump @swiftone - can we get this merged? Thanks!

PR #444, opened 2019-05-13
Require "https://" redirectUri unless running on localhost or testing.disableHttpsCheck

aarongranick-okta:
## PR Checklist

- [ Y ] The commit message follows our [guidelines](/okta/okta-oidc-js/blob/master/CONTRIBUTING.md#commit)
- [ Y ] Tests for the changes have been added (for bug fixes / features)
- [ Y ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ X ] Feature
- [ X ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [  X ] Documentation changes

Insecure `http://` urls are allowed for `redirectUri`. 

## What is the new behavior?
Secure `https://` urls will be required for `redirectUri` unless:
- `testing.disableHttpsCheck` - this will also disable secure cookie setting
- Running on `http://localhost` - specific exemption for local development

- [ X ] Yes
- [  ] No

With this change, an exception will be thrown if using http:// for redirectUri or appBaseUrl. To allow insecure cookies (and http:// urls), `testing.disableHttpsCheck` can be set to true.  

`http://localhost` is explicitly allowed to support local development and our sample apps. However, with a "secure" cookie setting, cookies will not be stored.

This would affect applications which expect to run in production on the insecure http:// protocol. With these changes, they must explicitly set `disableHttpsCheck`. A warning will still be shown in the developer console.

## Reviewers
@swiftone

### DIFF
--- a/packages/configuration-validation/src/lib.js
+++ b/packages/configuration-validation/src/lib.js
@@ -22,6 +22,7 @@ const copyCredentialsMessage = 'You can copy it from the Okta Developer Console 
   'in the details for the Application you created. ' +
   `Follow these instructions to find it: ${findAppCredentialsURL}`;
 
+const isLocal = new RegExp('^http://localhost');
 const isHttps = new RegExp('^https://');
 const hasProtocol = new RegExp('://');
 const hasDomainAdmin = /-admin.(okta|oktapreview|okta-emea).com/;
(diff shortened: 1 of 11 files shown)


packages/configuration-validation/src/lib.js
```diff
@@ -49,14 +50,16 @@ configUtil.buildConfigObject = (config) => {
   let tokenManager = config.tokenManager;
   const autoRenew = config.autoRenew || config.auto_renew;
   const storage = config.storage;
-  if (storage || autoRenew) {
+  const disableHttpsCheck = config.testing ? config.testing.disableHttpsCheck : false;
+  if (storage || autoRenew || disableHttpsCheck) {
     // Properties already defined within the "tokenManager" section will not be overwritten
```
swiftone: If any one of them is defined,  ALL of them will not be set?  Is that intended

aarongranick-okta: This logic is for creating a `tokenManager` section if it does not exist. If none of these are defined, we will not create an empty object.

aarongranick-okta:
Delayed until AuthJS 3.0

PR #453, opened 2019-05-17
Adds support for legacy tests on bacon

vijetmahabaleshwar-okta:
* We recently updated chrome driver to 2.46 as travis chrome version was upgraded
* This is causing our PDV tests to fail on bacon 
* Added scripts to take care of legacy chrome driver for bacon

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/okta-vue/test/e2e/harness/scripts/downloadChromeDriver.sh
+++ b/packages/okta-vue/test/e2e/harness/scripts/downloadChromeDriver.sh
@@ -1,18 +1,25 @@
 # Install ChromeDriver.
 OS="`uname`"
 
+if [ $# -eq 0 ]
+then
+  CHROME_DRIVER_VERSION=2.46
+else
+  CHROME_DRIVER_VERSION=2.45
+fi
+
 case $OS in
   'Linux')
     OS='Linux'
-    CHROME_DRIVER=http://chromedriver.storage.googleapis.com/2.46/chromedriver_linux64.zip
+    CHROME_DRIVER=http://chromedriver.storage.googleapis.com/${CHROME_DRIVER_VERSION}/chromedriver_linux64.zip
     ;;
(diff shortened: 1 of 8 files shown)


packages/okta-vue/test/e2e/harness/scripts/downloadChromeDriver.sh
```diff
@@ -1,18 +1,25 @@
 # Install ChromeDriver.
 OS="`uname`"
 
+if [ $# -eq 0 ]
```
robertjd: I'd feel better if we just tested for eq "legacy", then this code is more self-documenting

aarongranick-okta: Maybe even better, can pass in the version.  `./downloadChromeDriver 2.45`

aarongranick-okta:
Instead of duplicating tasks, it would be nice if we could run the same tasks with different chrome versions. What if we want to test with the next version? Are we going to add `test:e2e:next` and `pretest:e2e:next` ? Ideally this could all be controlled by an environment variable.

vijetmahabaleshwar-okta:
@aarongranick-okta - Even I don't like duplicating scripts but this solution makes it a bit simpler, I feel. Passing arguments to scripts would require more changes. Also, we would need only 2 versions of chrome driver - the latest, and legacy one for chrome. 
@wayneearl-okta is also working on script that automatically downloads the right chrome driver depending on the version of chrome. Once that is merged (at a later point), this will be much cleaner. For now, this is a quick fix to get our critical PDV tests passing.

PR #459, opened 2019-05-21
Wje okta 225391 remove chromedriver pinning

wayneearl-okta:
## PR Checklist

## PR Type
- [x] Refactoring Tests

- [ ] Yes
- [x] No

Add dynamic chromedriver update based on chrome version available in current environment


packages/okta-react/test/e2e/harness/e2e/page-objects/app.po.js
```diff
@@ -20,15 +20,15 @@ export class AppPage {
   waitUntilVisible() {
     const loginExists = ExpectedConditions.presenceOf(this.getLoginButton());
     const logoutExists = ExpectedConditions.presenceOf(this.getLogoutButton());
-    browser.wait(ExpectedConditions.or(loginExists, logoutExists), 5000);
+    browser.wait(ExpectedConditions.or(loginExists, logoutExists), 15000);
```
swiftone: Would it make sense to pull all these 15000 into a single constant that can be moved up/down as we deem reasonable?

wayneearl-okta: yes. it would. i was being lazy (like the last guy) ;)

packages/okta-vue/test/e2e/harness/test/e2e/nightwatch.conf.js
```diff
@@ -8,11 +8,13 @@ module.exports = {
 
   selenium: {
     start_process: true,
-    server_path: require('selenium-server').path,
+    // server_path: require('selenium-server').path,
```
swiftone: Slate comment here and a few lines below?

scripts/updateSeDrivers.sh
```diff
@@ -0,0 +1,102 @@
...
+fi
+
+if [[ -z ${SE_STANDALONE_VER} ]];
+then
+  SE_STANDALONE_VER=3.141.59
+fi
+
+if [[ -z ${CHROME_DRIVER_VER} ]];
+then
+    # MacOS or Linux?
```
swiftone: Does it make sense to pull the OS decision and the Chrome version decision into functions we can call from here?

scripts/updateSeDrivers.sh
```diff
@@ -0,0 +1,102 @@
...
+
+    # chrome version
+    if [[ ${OS} == "mac" ]];
+    then
+      TEMP_CHROME_VER=$(/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --version)
+    else
+      TEMP_CHROME_VER=$(google-chrome --product-version)
+    fi
+
+    echo "Chrome Version: ${TEMP_CHROME_VER}"
```
swiftone: Is this line still relevant?

packages/okta-vue/test/e2e/harness/test/e2e/nightwatch.conf.js
wayneearl-okta: oh. duh. comment should go.

scripts/updateSeDrivers.sh
wayneearl-okta: hmm. good question. i'll take a look.

wayneearl-okta: refactored script to be cleaner and less rambly

packages/okta-angular/test/e2e/harness/package.json
```diff
@@ -7,13 +7,11 @@
     "start": "yarn env && ng serve --port=3000",
     "env": "node ./scripts/prebuild.js",
     "build": "ng build",
-    "pretest": "./node_modules/.bin/webdriver-manager update --versions.chrome 2.46 --gecko false",
+    "pretest": "../../../../../scripts/updateSeDrivers.sh 0",
```
vijetmahabaleshwar-okta: Why do we need the "0" argument here?

packages/okta-react/test/e2e/harness/e2e/page-objects/app.po.js
vijetmahabaleshwar-okta: Is it better to have a global timeout variable? 
Also, any reason a timeout of 5000 doesn't work?

wayneearl-okta: Actually, there is a global timeout, set in the protractor config file. i think this is passing a shorter one, for no reason i can tell. i ended up refactoring it and getting rid of the override anyway.

packages/okta-angular/test/e2e/harness/package.json
wayneearl-okta: This was a sneaky one. i suppose i could have added a comment ... somewhere ... to explain it. It has to do with the 'workspace' configuration of the repo. some of the 'packages' depend on the `webdriver-manager` at the root level, and some depend on the one that's in the package. i'm passing in the extra param to tell it to use the local one.

PR #473, opened 2019-06-11
test(jwt-verifier): Adds debugging output for flaky test

swiftone:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [X] CI related changes
- [ ] Documentation changes

On rare-but-too-common occasions the test will fail with `jwtString.split is not a function`.

Let's figure out what jwtString is then, because it REALLY should be a string.  

This failure only occurs in CI testing, not locally.

Issue Number: N/A

- [ ] Yes
- [X] No


robertjd:
If it's not a string, doest the assertion lib print out the actual string value for us to inspect?

swiftone:
> If it's not a string, doest the assertion lib print out the actual string value for us to inspect?

It prints the toString() of it, yes - I wanted to capture a bit more info in case the toString() was misleading.

Issue #367, opened 2019-01-08
JwtVerifier -- Why assert clientId?


chadjaros:
## I'm submitting this issue for the package(s):

- [x] jwt-verifier
- [ ] okta-angular
- [ ] oidc-middleware
- [ ] okta-react
- [ ] okta-react-native
- [ ] okta-vue

## I'm submitting a:

- [ ] Bug report  
- [ ] Feature request
- [x] Question

## Current behavior

In version 0.13 clientId was added as a required field. 

## Expected behavior

Requiring clientID seems unnecessary. Why was this added as a requirement, rather than being handled similarly to audience where this claim is validated only if the argument is supplied?

Please note: Audience is technically required to be validated by the JWT spec if present in the token.

## Minimal reproduction of the problem with instructions

Create instance of JwtVerifier without supplying clientId property

## Extra information about the use case/user story you are trying to implement

Our servers have historically validated tokens from any client id. This provides us with a lot of flexibility as far as managing token behavior between multiple services.
```
[ UX 1 (Okta Application) ]   [ UX 2 (Okta Application)]
                         \         /
                           [Server]
```
## Environment

- Package Version: 0.14
- Node version (`node -v`): 8

swiftone:
Excellent point!  While clientId is required to validate an id token, it should not be required to validate an access token.  We are looking at the necessary changes involved, and will update this issue as progress is made.  In the meantime, it may be possible to use multiple instances of the verifier to resolve any conflicts you are encountering.

chadjaros:
In the mean time, we have migrated away from this library.

PR #477, opened 2019-06-17
feat(jwt-verifier): Removes check of client_id from access tokens

swiftone:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [X] Bugfix
- [X] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [X] Documentation changes

clientId/client_id was required, even for access tokens that, per spec, do not require client id

Issue Number: #367

## What is the new behavior?
clientId is now optional, and will only be asserted if such an assertion is requested via customClaims

- [ ] Yes
- [X] No

## Reviewers
@aarongranick-okta

### DIFF
--- a/packages/jwt-verifier/lib.js
+++ b/packages/jwt-verifier/lib.js
@@ -63,9 +63,10 @@ class OktaJwtVerifier {
   constructor(options = {}) {
     // Assert configuration
     assertIssuer(options.issuer, options.testing);
-    assertClientId(options.clientId);
+    if( options.clientId ) { 
+      assertClientId(options.clientId);
+    }
 
-    this.clientId = options.clientId;
     this.claimsToAssert = options.assertClaims || {};
     this.jwksClient = jwksClient({
       jwksUri: options.issuer + '/v1/keys',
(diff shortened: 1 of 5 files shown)


packages/jwt-verifier/lib.js
```diff
@@ -63,9 +63,10 @@ class OktaJwtVerifier {
   constructor(options = {}) {
     // Assert configuration
     assertIssuer(options.issuer, options.testing);
-    assertClientId(options.clientId);
+    if( options.clientId ) {
```
robertjd: Do we even need `assertClientId` anymore? Or should we just let this be handled by the custom claims assertion logic?

packages/jwt-verifier/test/spec/configuration.spec.js
```diff
@@ -95,22 +94,14 @@ describe('jwt-verifier configuration validation', () => {
     expect(createInstance).toThrow();
   });
 
-  it('should throw if the client_id is not provided', () => {
+  it('should NOT throw if client_id not matching {clientId} is provided', () => {
```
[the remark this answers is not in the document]
swiftone: Asserting that the input is allowed, and is NOT checked against the claims automatically.

That's not exciting, but this documents the expected behavior for this case.

packages/jwt-verifier/lib.js
swiftone: The value here is catching "{clientId}" cases.

PR #482, opened 2019-06-26
Reworked React Native SDK

kitty7756:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [x] Other... Please describe: Rework of the entire sdk

Current React Native SDK relies on Unimodules and Expo modules. Even with this dependency, there is still a security issue on android devices where launching the browser follows MainAppActivity ->Browser -> MainAppActivity flow while the correct flow should be MainAppActivity -> OktaAuthenticationActivity -> Browser -> RedirectActivity -> OktaAuthenticationActivity ->
MainAppActivity.

## What is the new behavior?
To solve the above problems, this pull request introduces a completely reworked React Native 
SDK that acts as a bridge for communicating with okta-oidc-android sdk and okta-oidc-ios sdk. These SDKs have been around for a while and they currently follow the best practice for native apps. This PR also completely gets rid of Unimodules and Expo dependencies. 

- [x] Yes
- [] No

This PR will remove the prerequisite of using Expo. For existing users who are currently using Expo, they have to eject from Expo in order to use the new API. Existing users will also have to consume the API differently. For usage, refer to usage in README.md


swiftone:
> This PR does not introduce a breaking change to any of the existing code in the repo.

Does this require a change from anyone consuming the API?  Including those that were using Expo in certain ways?

packages/okta-react-native/README.md
```diff
@@ -32,12 +25,9 @@ In Okta, applications are OpenID Connect clients that can use Okta Authorization
...
 | ------------------- | -------------------------------------------- |
-| App Name            | My Native App                                |
-| Login redirect URIs | com.oktapreview.{yourOrg}:/+expo-auth-session|
-|                     | exp://localhost:19000/+expo-auth-session     |
-| Grant Types Allowed | Authorization Code, Refresh Token            |
-
-**Note:** *Login redirect URIs will depend on your scheme. For setting up a scheme on Android and iOS, refer to [configuration](#configuration).*
+| App Name            | {YOUR APP NAME}                              |
+| Login redirect URIs | {YOUR_REDIRECT_URL}                          |
+| Grant Types Allowed | {PICK_YOUR_SCOPES}                           |
```
swiftone: Are these terms defined anywhere?  Do we have a doc page we can point a user to where they can figure out what these values would be? 

I'm weak on mobile, but would https://developer.okta.com/docs/guides/sign-into-mobile-app/ be a helpful thing, or does that not cover these values?

packages/okta-react-native/README.md
```diff
@@ -50,145 +40,218 @@ After you have created the application there are two more values you will need t
...
-* `authorization_endpoint` - override the authorization endpoint from the well-known endpoint
-* `storageKey` - Unique key used to store/retrieve secure data
-* `keychainService` - See [Expo - keychainService](https://docs.expo.io/versions/latest/sdk/securestore.html#keychainservice-string-)
-* `keychainAccessible` - (iOS only) Specify when the stored item is accessible. See [Expo - keychainAccessible](https://docs.expo.io/versions/latest/sdk/securestore.html#keychainaccessible-enum-)
+This method will create a configured client on the native modules.
 
-### `signInWithRedirect`
+```javascript
+await createConfig(
+	'<YOUR_CLIENT_ID>',
```
swiftone: elsewhere we have a fairly standardized format for these using `{}` (e.g. `{clientId}`).  Is there a mobile standard to use `<>`?

packages/okta-react-native/index.js
```diff
@@ -0,0 +1,94 @@
+import { NativeModules, Platform } from 'react-native';
+import {
+	assertIssuer,
+  assertClientId,
+	assertRedirectUri
```
swiftone: Watch out for mixing tabs and spaces (as shown here, it can look fine in one editor and get crazy in another).  JS community has largely standardized on 2 spaces for indent and you can set your editor to translate a TAB into the 2 spaces.

packages/okta-react-native/index.js
```diff
@@ -0,0 +1,94 @@
...
+} from '@okta/configuration-validation';
+import jwt from 'jwt-lite';
+
+export const createConfig = async (
+  clientId,
+  redirectUri, 
+  endSessionRedirectUri, 
+  discoveryUri,
+  scopes) => {
```
swiftone: That is a very long function signature.  Have you considered using object destructuring?  
```
export const createConfig = async ({ 
  clientId, 
  redirectUri, 
  endSessionRedirectUri, 
  discoveryUri,
  scopes
}) => { 
```

That would require changing how this function is called (you pass an object of all the params rather than just the params) but has 2 key advantages:
* The order of the params no longer matters, as they are identified by key name rather than position
* Each argument is essentially named by the key, making reading the code more clear, particularly with many function arguments.
e.g. `createConfig({ clientId: 'DFG352SF2SR', redirectUri: 'http://myapp.com/', // etc })`

Object property name shorthand means that when variable names match up to the property names this can be less tedious:

```
const clientId = '123456';
const redirectUri = 'http://myapp.com/';
createConfig({ clientId, redirectUri, //etc}); 
// same as
// createConfig({ clientId: clientId, redirectUri: redirectUri //etc })
```

Object destructuring for function params may not always be the best choice, but I find I always want it when I have 3 or more params.

packages/okta-react-native/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+  assertIssuer(discoveryUri);
+  assertClientId(clientId);
+  assertRedirectUri(redirectUri);
+  assertRedirectUri(endSessionRedirectUri);
+
+  if (Platform.OS === 'ios') {
+		scopes = scopes.join();
```
swiftone: Tabs again.

packages/okta-react-native/index.js
```diff
@@ -0,0 +1,94 @@
...
+		scopes = scopes.replace(/,/g, ' ');
```
swiftone: `.join()` can take an argument for the separator, which defaults to ','.  If you said
`scopes = scopes.join(' ');` above, this line would be unnecessary.

packages/okta-react-native/index.js
```diff
@@ -0,0 +1,94 @@
...
+  return NativeModules.OktaSdkBridge.getUser();
+}
+
+export const getUserFromIdToken = async() => {
+  try {
+    let idTokenResponse = await getIdToken();
+    return jwt.decode(idTokenResponse.id_token).claimsSet;
+  } catch(error) {
+    throw error;
+  }
```
swiftone: Why the try/catch if the behavior will be the same as without the try catch?

packages/okta-react-native/package.json
```diff
@@ -1,59 +1,44 @@
...
-    "browser": true,
-    "preset": "jest-expo",
-    "automock": false,
-    "setupFiles": [
-      "./setupJest.js"
-    ],
-    "transformIgnorePatterns": [
-      "node_modules/(?!(jest-)?react-native|react-clone-referenced-element|@react-native-community|expo(nent)?|react-navigation|@react-navigation/.*|@unimodules/.*|sentry-expo|native-base)"
-    ]
+    "test": "echo \"Error: no test specified\"",
```
swiftone: Do we really have no way to run tests?

packages/okta-react-native/package.json
```diff
@@ -1,59 +1,44 @@
...
+    "oidc"
+  ],
+  "license": "Apache-2.0",
+  "readmeFilename": "README.md",
   "peerDependencies": {
-    "react-native": "*",
-    "react-native-securerandom": "^0.1.1",
-    "react-native-unimodules": "0.3.1"
+    "react": "^16.5.0",
+    "react-native": ">=0.57.0-rc.0 <1.0.x"
```
swiftone: A 1.0 is a major version on our side - is pre-1.0.x react-native something we want to support in our 1.x version?  If we later stop supporting it, we'd have to bump our major version.  (I know nothing of how much 0.57+ is used in react-native, so the answer may just be "yes")

kitty7756: The latest react-native version is 0.60c (which is still in beta). I guess for now we would want to support versions from 0.57.0 - 0.59.9 (the most stable versions)

packages/okta-react-native/package.json
kitty7756: Tests isn't a part of this PR. They will be introduced later in the next few weeks!

packages/okta-react-native/index.js
kitty7756: I was thinking to catch the error from `let idTokenResponse = await getIdToken();`, would that be the same with removing try and catch?

swiftone: You do catch it, but then immediately throw it when caught.  So with or without the try/catch, the error will be thrown.

packages/okta-react-native/index.js
kitty7756: that does sound good to me!

packages/okta-react-native/README.md
kitty7756: These aren't defined, the closest I found is https://developer.okta.com/docs/guides/sign-into-mobile-app/android/create-okta-application/, but this is only for android setup. maybe we should introduce a setup for react native only?

packages/okta-react-native/README.md
kitty7756: i'll make sure to change into the standardized format

packages/okta-react-native/README.md
swiftone: Probably, I'd say create a JIRA ticket (leave it unassigned) and Rob/Nate/Albert will prioritize it.  (It is probably chicken/egg waiting on these changes)

Meanwhile, for this PR, you should give at least a little guidance on how to figure out what these values are - either a place in the docs to look, a place in the the okta dashboard to look, or a sample value.  That reduces developer frustration when trying to follow your README when they aren't super-familiar with using Okta.

packages/okta-react-native/package.json
swiftone: 👍

packages/okta-react-native/package.json
swiftone: Good to know.  Yes, we don't want to advertise compatibility with anything that may break the API (which sub 1.0 versions can do anytime).  Normally this means we stay in the same major version, but if react-native isn't 1.0 yet, we should limit to the most recent version we've tested against, which can be the beta or the stable, whichever you feel more confident about.

packages/okta-react-native/OktaSdkBridgeReactNative.podspec
```diff
@@ -0,0 +1,20 @@
...
+  s.name         = package['name']
+  s.version      = package['version']
+  s.summary      = package['description']
+  s.license      = package['license']
+
+  s.authors      = package['author']
+  s.homepage     = package['homepage']
+  s.platform     = :ios, "9.0"
+
+  s.source       = { :git => "https://github.com/FormidableLabs/react-native-app-auth.git" }
```
IldarAbdullin-okta: FormidableLabs? :)  Please change it to Okta link

packages/okta-react-native/OktaSdkBridgeReactNative.podspec
```diff
@@ -0,0 +1,20 @@
...
+
+Pod::Spec.new do |s|
+  s.name         = package['name']
+  s.version      = package['version']
+  s.summary      = package['description']
+  s.license      = package['license']
+
+  s.authors      = package['author']
+  s.homepage     = package['homepage']
+  s.platform     = :ios, "9.0"
```
IldarAbdullin-okta: OIDC lib targets iOS11+. Let's change to 11.0

packages/okta-react-native/README.md
```diff
@@ -50,145 +40,219 @@ After you have created the application there are two more values you will need t
...
+
+   1. Create an XCode Workspace.
+   2. Add `okta-oidc.xcodeproj` to your Workspace.
+   3. Include `OktaOidc.framework` as a linked framework for your target (in the "General -> Linked Framework and
+      Libraries" section of your target).
+   4. Add `okta-oidc/okta-oidc/Okta` to your search paths of your target ("Build Settings -> "Header Search
+      Paths").
+
+#### Swift Configuration
+Since React Native uses Objective-C, and Okta React Native library is a swift wrapper, you will need to have at least one Swift file in your iOS project for the project to compile. To do this, follow the following steps:
```
IldarAbdullin-okta: Fairly speaking you don't need to add dummy swift file. I would write something like:
`Add project setting SWIFT_VERSION = 4.2. To do this open Build Settings in Xcode, select Edit -> Add Build setting -> Add User-Defined Setting. Specify SWIFT_VERSION and 4.2 as setting name and value correspondently.`

packages/okta-react-native/index.js
```diff
@@ -0,0 +1,86 @@
...
+  return NativeModules.OktaSdkBridge.createConfig(
+    clientId,
+    redirectUri,
+    endSessionRedirectUri,
+    discoveryUri,
+    scopes
+  );
+} 
+
+export const signIn = async () => {
```
IldarAbdullin-okta: nit style: check for the space in `async ()` declaration, sometimes there is no space between async and ()

packages/okta-react-native/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+export const getIdToken = async () => {
+  return NativeModules.OktaSdkBridge.getIdToken();
+}
+
+export const getUser = async () => {
+  return NativeModules.OktaSdkBridge.getUser();
+}
+
+export const getUserFromIdToken = async() => {
```
IldarAbdullin-okta: Is `async` really required for the fast operations? Can it be just blocking calls?

packages/okta-react-native/ios/OktaSDKError.swift
```diff
@@ -0,0 +1,67 @@
...
+            return NSLocalizedString("Id token does not exist", comment: "")
+        case .oktaOidcError:
+            return NSLocalizedString("Okta Oidc error", comment: "")
+        case .errorTokenType:
+            return NSLocalizedString("Token type not found", comment: "")
+        case .errorPayload:
+            return NSLocalizedString("Error in retrieving payload", comment: "")
+        }
+    }
+    public var errorCode: String? {
```
IldarAbdullin-okta: I've never seen error codes as `String`, can it be Int?

packages/okta-react-native/ios/OktaSdkBridge.podspec
```diff
@@ -0,0 +1,22 @@
...
+  s.summary      = "ROktaSdkBridgeNNewlibrary"
+  s.description  = <<-DESC
+                  RNNewlibrary
+                   DESC
+  s.homepage     = ""
+  s.license      = "MIT"
+  # s.license      = { :type => "MIT", :file => "FILE_LICENSE" }
+  s.author             = { "author" => "author@domain.cn" }
+  s.platform     = :ios, "9.0"
+  s.source       = { :git => "https://github.com/author/OktaSdkBridge.git", :tag => "master" }
```
IldarAbdullin-okta: Check the link please

packages/okta-react-native/ios/OktaSdkBridge.podspec
```diff
@@ -0,0 +1,22 @@
...
+  s.version      = "1.0.0"
+  s.summary      = "ROktaSdkBridgeNNewlibrary"
+  s.description  = <<-DESC
+                  RNNewlibrary
+                   DESC
+  s.homepage     = ""
+  s.license      = "MIT"
+  # s.license      = { :type => "MIT", :file => "FILE_LICENSE" }
+  s.author             = { "author" => "author@domain.cn" }
+  s.platform     = :ios, "9.0"
```
IldarAbdullin-okta: 11.0

packages/okta-react-native/ios/OktaSdkBridge.podspec
```diff
@@ -0,0 +1,22 @@
+
+Pod::Spec.new do |s|
+  s.name         = "OktaSdkBridge"
+  s.version      = "1.0.0"
+  s.summary      = "ROktaSdkBridgeNNewlibrary"
```
IldarAbdullin-okta: Just `ROktaSdkBridge`

BTW, should it contain `RN` prefix, so it shows that it is react native lib. Something like RNOktaOidcBridge

packages/okta-react-native/ios/OktaSdkBridge.podspec
```diff
@@ -0,0 +1,22 @@
...
+  s.description  = <<-DESC
+                  RNNewlibrary
```
IldarAbdullin-okta: Just ROktaSdkBridge

packages/okta-react-native/ios/OktaSdkBridge.podspec
```diff
@@ -0,0 +1,22 @@
...
+                   DESC
+  s.homepage     = ""
```
IldarAbdullin-okta: Link to the git with xcode project

packages/okta-react-native/ios/OktaSdkBridge.podspec
```diff
@@ -0,0 +1,22 @@
...
+  s.license      = "MIT"
+  # s.license      = { :type => "MIT", :file => "FILE_LICENSE" }
+  s.author             = { "author" => "author@domain.cn" }
```
IldarAbdullin-okta: your email

packages/okta-react-native/ios/OktaSdkBridge.swift
```diff
@@ -0,0 +1,373 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import Foundation
+import OktaOidc
+
+@objc(OktaSdkBridge)
```
IldarAbdullin-okta: OktaSdkBridge is too common class name. In fact it is OIDC bridge between js and native code
I would name it something like: RNOktaOidcBridge

packages/okta-react-native/ios/OktaSdkBridge.swift
```diff
@@ -0,0 +1,373 @@
...
+            promiseResolver(true)
+        } catch let error {
+            promiseRejecter(OktaReactNativeError.oktaOidcError.errorCode, error.localizedDescription, error)
+        }
+    }
+    
+    @objc(signIn:promiseRejecter:)
+    func signIn(_ promiseResolver: @escaping RCTPromiseResolveBlock, promiseRejecter: @escaping RCTPromiseRejectBlock) {
+        
+        let view = RCTPresentedViewController();
```
IldarAbdullin-okta: Move it after the guard checks

packages/okta-react-native/ios/OktaSdkBridge.swift
```diff
@@ -0,0 +1,373 @@
...
+        
+        guard let currOktaOidc = oktaOidc else {
+            let error = OktaReactNativeError.notConfigured
+            promiseRejecter(error.errorCode, error.errorDescription, error)
+            return
+        }
+        
+        guard let currView = view else {
```
IldarAbdullin-okta: `guard let view = RCTPresentedViewController() else `

packages/okta-react-native/ios/OktaSdkBridge.swift
```diff
@@ -0,0 +1,373 @@
...
+                OktaSdkConstant.RESOLVE_TYPE_KEY: OktaSdkConstant.AUTHORIZED,
+                OktaSdkConstant.ACCESS_TOKEN_KEY: stateManager?.accessToken
+            ]
+            
+            promiseResolver(dic)
+        }
+    }
+    
+    @objc(signOut:promiseRejecter:)
+    func signOut(_ promiseResolver: @escaping RCTPromiseResolveBlock, promiseRejecter: @escaping RCTPromiseRejectBlock) {
```
IldarAbdullin-okta: Same comments as for the `SignIn` part

packages/okta-react-native/ios/OktaSdkBridge.swift
```diff
@@ -0,0 +1,373 @@
...
+        }
+        
+        guard let stateManager = OktaOidcStateManager.readFromSecureStorage(for: oidcConfig) else {
+            let error = OktaReactNativeError.unauthenticated
+            promiseRejecter(error.errorCode, error.errorDescription, error)
+            return
+        }
+        
+        let dic = [
+            OktaSdkConstant.ACCESS_TOKEN_KEY: stateManager.accessToken
```
IldarAbdullin-okta: `accessToken ` is optional. Do you need additional guard?

packages/okta-react-native/OktaSdkBridgeReactNative.podspec
kitty7756: Whoops :D

packages/okta-react-native/ios/OktaSDKError.swift
kitty7756: I thought it was supposed int at first, but apparently when I pass that in in PromiseRejecter it tells me that the code has to be a string :s weird!

packages/okta-react-native/README.md
kitty7756: I have to test this to verify, will update the doc once it's being tested!

packages/okta-react-native/index.js
kitty7756: Thanks for checking this, I deleted all the space to keep it consistent

packages/okta-react-native/index.js
kitty7756: i'm using async here since the function still has to get the id token from the native modules, and that is an async operation :p

packages/okta-react-native/ios/OktaSdkBridge.podspec
kitty7756: Would that change affect all the module names? My current module name is named to OktaSdkBridge

packages/okta-react-native/ios/OktaSdkBridge.podspec
kitty7756: Do we want to keep the author field? I'm deleting it since I don't think its needed

packages/okta-react-native/ios/OktaSdkBridge.swift
kitty7756: Will do this later since that might be a breaking change and i would want everything else to work first

packages/okta-react-native/ios/OktaSdkBridge.swift
kitty7756: additional guard added

packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
```diff
@@ -0,0 +1,457 @@
...
+                    .endSessionRedirectUri(endSessionRedirectUri)
+                    .scopes(scopeArray)
+                    .discoveryUri(discoveryUri)
+                    .create();
+
+            this.webClient = new Okta.WebAuthBuilder()
+                    .withConfig(config)
+                    .withContext(reactContext)
+                    .withStorage(new SharedPreferenceStorage(reactContext))
+                    .setRequireHardwareBackedKeyStore(true) // TODO: remember to set it to true when releasing SDK
```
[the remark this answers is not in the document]
kitty7756: how do we want the keystore configurable?

packages/okta-react-native/README.md
kitty7756: I added some sample values, let me know if it looks good

packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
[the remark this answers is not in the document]
FeiChen-okta: can pass a boolean parameter instead of hardcoding setRequireHardwareBackedKeyStore(true).

packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
```diff
@@ -0,0 +1,462 @@
...
+
+            if (webClient == null) {
+                promise.reject(OktaSdkError.NOT_CONFIGURED.getErrorCode(), OktaSdkError.NOT_CONFIGURED.getErrorMessage());
+                return;
+            }
+
+            final SessionClient sessionClient = webClient.getSessionClient();
+            final WritableMap params = Arguments.createMap();
+
+            webClient.registerCallback(new ResultCallback<AuthorizationStatus, AuthorizationException>() {
```
FeiChen-okta: The callback can be triggered without the
 
```java
webClient.signIn(currentActivity, null);
``` 
being called. An edge case where the ReactActivity -> WebBrowser. ReactActivity is destroyed so the callback is broken. But the OIDC SDK will keep track of the browser results so whenever the callback is registered again, the status is returned immediately. So will webClient.signIn be called if the promise is resolved?

packages/okta-react-native/index.js
IldarAbdullin-okta: Just thinking out loud.  I don't know too much about js world, but why we just can't get all required data once from native code and then just keep it in js part. I mean we can serialize access token, id token and etc. and return to js part during `signIn` call. Then we can get rid of async bridging functions, like getIdToken and getAccessToken

kitty7756: That's a good idea to to think about, but I also had concerns with security if we take this approach. Since on native modules the tokens are stored securely. If we expose it to JS then JS has no way to store those securely.

packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
kitty7756: Resolved

packages/okta-react-native/android/build.gradle
```diff
@@ -0,0 +1,140 @@
...
+  compileSdkVersion safeExtGet('compileSdkVersion', DEFAULT_COMPILE_SDK_VERSION)
+  buildToolsVersion safeExtGet('buildToolsVersion', DEFAULT_BUILD_TOOLS_VERSION)
+
+  defaultConfig {
+      minSdkVersion safeExtGet('minSdkVersion', DEFAULT_MIN_SDK_VERSION)
+      targetSdkVersion safeExtGet('targetSdkVersion', DEFAULT_TARGET_SDK_VERSION)
+      versionCode 1
+      versionName "1.0"
+      manifestPlaceholders = [
+              "appAuthRedirectScheme": "com.oktareactnative.samples-test"
```
Fei-Chen-: Leave this blank.

packages/okta-react-native/README.md
```diff
@@ -50,145 +40,387 @@ After you have created the application there are two more values you will need t
...
+#### Install Okta Open ID Connect iOS
+This library depends on the native [Okta OIDC iOS](https://github.com/okta/okta-oidc-ios) library. This library is not distributed as part of the React Native library to keep your dependency management consistent. 
+
+There are three ways to add Okta OIDC iOS to your dependencies:
+
+1. **CocoaPods**
+   With [CocoaPods](https://guides.cocoapods.org/using/getting-started.html), add the following line to
+   your `Podfile`:
+       
+		pod 'OktaOidc'
```
IldarAbdullin-okta: Please add optimistic operator to your example with 3.0 version. We should show that current SDK compatible only with 3.x.x iOS OIDC versions
`pod 'OktaOidc', '~> 3.0'`

packages/okta-react-native/README.md
```diff
@@ -50,145 +40,392 @@ After you have created the application there are two more values you will need t
...
+This library depends on the native [Okta OIDC iOS](https://github.com/okta/okta-oidc-ios) library. This library is not distributed as part of the React Native library to keep your dependency management consistent. 
+
+There are three ways to add Okta OIDC iOS to your dependencies:
+
+1. **CocoaPods**
+   With [CocoaPods](https://guides.cocoapods.org/using/getting-started.html), add the following line to
+   your `Podfile`:
+
+    ```   
+    pod 'OktaOidc', '~> 3.0'
```
IldarAbdullin-okta: I would show the whole podfile example. `use_frameworks` is also critical command in podfile
So, something like that:

`platform :ios, '11.0'

use_frameworks!

target 'YourTarget' do
pod 'OktaOidc', '~> 3.0'
end`

packages/okta-react-native/README.md
```diff
@@ -50,145 +40,404 @@ After you have created the application there are two more values you will need t
...
+
+There are two ways to add Okta OIDC iOS to your dependencies:
+
+1. **CocoaPods**
+   With [CocoaPods](https://guides.cocoapods.org/using/getting-started.html), make sure your `Podfile` looks like this:
+
+    ```   
+    platform :ios, '11.0'
+
+    target 'Test9OldVersion' do
```
IldarAbdullin-okta: Please rename `Test9OldVersion` to `{YourTargetName}`

packages/okta-react-native/README.md
```diff
@@ -50,145 +40,404 @@ After you have created the application there are two more values you will need t
...
+    github "okta/okta-oidc-ios"
+    ```
+
+   Then run `carthage update --platform iOS`.
+
+   Drag and drop `OktaOidc.framework` from `ios/Carthage/Build/iOS` to `Frameworks` in Xcode.
+
+   Add a copy files build step for `OktaOidc.framework`: open Build Phases on Xcode, add a new "Cope Files" phase, choose "Frameworks" as destination, add `OktaOidc.framework` and ensure "Code Sign on Copy" is checked. -->
+
+2. **Static Library**
```
IldarAbdullin-okta: I would delete this section. It causes a lot of confusion:
- we don't add any static libraries here
- this type of configuration requires some extra steps: git checkout of oidc repo with particular tag, then adding it as a dependency

I think pod installation should be enough for now. In next release we will add Carthage support

PR #518, opened 2019-08-02
Use lerna to manage monorepo packages

aarongranick-okta:
This gives separate yarn.lock files for each project under packages/ as well as the 3 test harness applications. There is no hoisting.

Removes the duplicate local copy of okta-angular & okta-react by test harness. Using webpack config for react app, using web_module shim for angular.

Contains build fixes for java & chrome version.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Yarn workspace, by default, hoists all dependencies to a top level node_modules folder. This causes all projects to share the same set of node_modules. This is desirable in a monorepo with many closely related modules who want to keep their dependency versions closely in sync.

This monorepo contains sdks that are loosely related to each other. We want to upgrade dependencies individually without affecting the others. Thus, the isolation of separate node_modules folders, allowing divergent versions, is what we want.

## What is the new behavior?

- There is now a yarn.lock for each module in packages folder
- Running `yarn install` within a child folder will not install all repo dependencies, but will act locally.

- [ ] Yes
- [x] No

## Other information
https://oktainc.atlassian.net/browse/OKTA-243117

## Reviewers

@swiftone 
@vijetmahabaleshwar-okta


packages/okta-react/test/e2e/harness/src/CustomLogin.js
```diff
@@ -14,8 +14,10 @@ import { Component } from 'react';
 import { withAuth } from '@okta/okta-react';
 
 export default withAuth(class CustomLogin extends Component {
-  render() {
+  componentDidMount() {
```
manueltanzi-okta: Not sure I understand why these changes are needed (wasn't expecting tests changes in this PR), how passing from yarn workspace to lerna changes test behavior?

packages/okta-react/test/e2e/harness/src/SessionTokenLogin.js
```diff
@@ -62,7 +63,6 @@ export default withAuth(class SessionTokenLogin extends Component {
 
   render() {
     if (this.state.sessionToken) {
-      this.props.auth.redirect({sessionToken: this.state.sessionToken});
```
manueltanzi-okta: ditto

packages/okta-react/test/e2e/harness/config-overrides.js
```diff
@@ -0,0 +1,21 @@
+/* global __dirname, module */
+
+var path = require('path');
```
manueltanzi-okta: nit: const

packages/okta-react/test/e2e/harness/src/CustomLogin.js
aarongranick-okta: Performing the redirect in render() is a problem if the component renders more than once.  I experienced this during debugging. I believe it will usually render twice, which causes the first redirect request to cancel.

packages/okta-react/test/e2e/harness/src/SessionTokenLogin.js
aarongranick-okta: same explanation. Unlike our courage library, React components will render whenever state / props are changed. when using our `withAuth` higher order component it will trigger render for us.  We want to redirect once and ONLY once.

packages/okta-react/test/e2e/harness/src/CustomLogin.js
aarongranick-okta: I believe this used to work in auth-js < 2.6 because `getWithRedirect` was synchronous. Now there it is async and there is time for the 2nd request to fire.

aarongranick-okta:
@manueltanzi-okta The reason there are test fixes in this PR is because it began with trying to fix a broken build due to test failures with the `@okta/okta-react` with recently upgraded `@okta/okta-auth-js`. There were additional changes necessary to the test harness app in order to be compatible with the latest version of okta-auth-js. However, tests in CI were passing because the top-level yarn.lock file / yarn hoisting was testing using an older version of okta-auth-js.

While working on the test harness I became blocked by 2 things:
- Local module `@okta/okta-react` was being copied to test harness node_modules. This made it very difficult to make changes that would affect the tests and led to a lot of confusion. I replaced this mechanism (file dependency in package.json) with a webpack config.
- Because of yarn workspace running `yarn install` in ANY directory (such as test harness) would perform a full install (and prepare/build) on the entire repo. This became a blocker for me as I was investigating/fixing the first issue (local module copy)
- Local module `@okta/okta-angular` had the same local module copy issue as `@okta/okta-react`, but was not as easy to fix. Custom webpack config for Angular 6 was complicated, so I used a shim in `web_modules` (webpack convention) to load the webpacked version and avoid copy.

PR #529, opened 2019-08-19
@okta/configuration-validation: buildConfigObject: support all config options

aarongranick-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: 244252

Previously only specific config values were passed to okta-auth-js

## What is the new behavior?

All config values are passed. Legacy support for underscore variables, and tokenManager config.

- [ ] Yes
- [x] No

## Other information

Implicitly adds support for the `pkce` config option

## Reviewers

@manueltanzi-okta

### DIFF
--- a/packages/configuration-validation/package.json
+++ b/packages/configuration-validation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@okta/configuration-validation",
-  "version": "0.3.0",
+  "version": "0.4.0",
   "description": "Configuration validation support for Okta JavaScript SDKs",
   "main": "./dist/lib.js",
   "files": [
@@ -35,6 +35,7 @@
     "babel-cli": "^6.26.0",
     "babel-eslint": "^8.1.1",
     "babel-preset-env": "^1.7.0",
+    "deep-extend": "^0.6.0",
     "eslint": "^4.7.1",
     "jest": "^23.6.0"
   },
--- a/packages/configuration-validation/package.json
+++ b/packages/configuration-validation/package.json
@@ -37,7 +37,8 @@
     "babel-preset-env": "^1.7.0",
     "deep-extend": "^0.6.0",
     "eslint": "^4.7.1",
-    "jest": "^23.6.0"
+    "jest": "^23.6.0",
+    "lodash": "^4.17.15"
   },
   "jest": {
     "testEnvironment": "jsdom"
--- a/packages/configuration-validation/test/lib.test.js
+++ b/packages/configuration-validation/test/lib.test.js
@@ -29,6 +29,8 @@ describe('Configuration Validation', () => {
         clientId: '{clientId}',
         issuer: '{issuer}',
         redirectUri: '{redirectUri}',
+        storage: '{storage}',
+        autoRenew: '{autoRenew}',
         tokenManager: {
           storage: '{storage}',
           autoRenew: '{autoRenew}'
@@ -47,11 +49,15 @@ describe('Configuration Validation', () => {
 
       expect(buildConfigObject(passedConfig)).toEqual({
         clientId: '{client_id}',
+        client_id: '{client_id}',
         issuer: '{issuer}',
         redirectUri: '{redirect_uri}',
+        redirect_uri: '{redirect_uri}',
+        storage: '{storage}',
+        auto_renew: '{auto_renew}',
         tokenManager: {
           storage: '{storage}',
-          autoRenew: '{auto_renew}'
+          autoRenew: '{auto_renew}',
         }
       });
     });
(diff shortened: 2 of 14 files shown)


packages/configuration-validation/yarn.lock
```diff
@@ -1159,6 +1159,7 @@ decode-uri-component@^0.2.0:
 deep-extend@^0.6.0:
   version "0.6.0"
   resolved "https://registry.yarnpkg.com/deep-extend/-/deep-extend-0.6.0.tgz#c4fa7c95404a17a9c3e8ca7e1537312b736330ac"
+  integrity sha512-LOHxIOaPYdHlJRtCQfDIVZtfw/ufM8+rVj649RIHzcm/vGwQRXFt6OPqIFWsm2XEMrNIEtWR64sY1LEKD2vAOA==
```
manueltanzi-okta: Is there a reason only this dependency has integrity sha?

packages/configuration-validation/package.json
```diff
@@ -35,8 +35,10 @@
     "babel-cli": "^6.26.0",
     "babel-eslint": "^8.1.1",
     "babel-preset-env": "^1.7.0",
+    "deep-extend": "^0.6.0",
     "eslint": "^4.7.1",
-    "jest": "^23.6.0"
+    "jest": "^23.6.0",
+    "lodash": "^4.17.15"
```
swiftone: Note for anyone curious: This lodash version addresses the known security issues.

packages/configuration-validation/test/lib.test.js
```diff
@@ -70,14 +76,76 @@ describe('Configuration Validation', () => {
...
+      const passedConfig = {
+        storage: '{storage}',
+        autoRenew: '{autoRenew}',
+        tokenManager: {
+          secure: true
+        }
+      }
+
+      expect(buildConfigObject(passedConfig)).toEqual({
+        storage: '{storage}',
```
manueltanzi-okta: `storage` and `autoRenew` should not belong on top-level but to `tokenManager` only

packages/configuration-validation/test/lib.test.js
```diff
@@ -70,14 +76,76 @@ describe('Configuration Validation', () => {
 
       expect(buildConfigObject(passedConfig)).toEqual({
         clientId: '{clientId}',
+        client_id: '{client_id}',
         issuer: '{issuer}',
         redirectUri: '{redirectUri}',
+        redirect_uri: '{redirect_uri}',
+        storage: '{storage}',
+        autoRenew: '{autoRenew}',
+        auto_renew: '{auto_renew}',
```
swiftone: Do we use `auto_renew` somewhere?  If it's deprecated, should we remove it from the test I/O in all but a "deprecated still works" test?

packages/configuration-validation/yarn.lock
```diff
@@ -2600,7 +2601,7 @@ lodash.sortby@^4.7.0:
   version "4.7.0"
   resolved "https://registry.yarnpkg.com/lodash.sortby/-/lodash.sortby-4.7.0.tgz#edd14c824e2cc9c1e0b0a1b42bb5210516a42438"
 
-lodash@^4.17.11, lodash@^4.17.14, lodash@^4.17.4, lodash@^4.2.0, lodash@^4.3.0:
+lodash@^4.17.11, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4, lodash@^4.2.0, lodash@^4.3.0:
```
swiftone: Why does this still have 4.17.4, etc?  There are some notable security issues on those, do we have some other deps we should update the minimum version on?

packages/okta-vue/yarn.lock
```diff
@@ -16,6 +16,10 @@
     esutils "^2.0.2"
     js-tokens "^4.0.0"
 
+"@okta/configuration-validation@^0.3.0":
```
swiftone: Why are these updating to 0.3.0 here (where you're creating 0.4.0)?

packages/configuration-validation/test/lib.test.js
aarongranick-okta: It has been in the documentation for awhile, especially on the react side, where it was actually required to use the underscore syntax until very recently. I think we will want to let it settle for awhile before removing

packages/okta-vue/yarn.lock
aarongranick-okta: I have left the versions untouched in the various packages. The first one to require 0.4 will be React in a subsequent PR. I can update everything to 0.4 if we prefer.

packages/configuration-validation/yarn.lock
aarongranick-okta: I am not sure, I will re-update the lock files and see if it changes

packages/configuration-validation/test/lib.test.js
swiftone: Makes sense - perhaps a comment here to say `// from okra-react vx.x ` or whenever, so we know when it can be removed?

packages/configuration-validation/test/lib.test.js
aarongranick-okta: I am preserving legacy behavior, where these specific parameters were hoisted down into the `tokenManager` section. The side-effect is that there are 'duplicate' values at the top-level, but this does not cause any negative behavior.

packages/okta-vue/yarn.lock
swiftone: I'm more curious why your changes to configuration-validation are making a change in the other packages at all

packages/configuration-validation/yarn.lock
aarongranick-okta: I'm not sure about this. I added the lodash dependency, so there should be nothing else that depends on it. I will clean and update yarn locks to see if this changes anything

packages/okta-vue/yarn.lock
aarongranick-okta: For some reason their reference to configuration-validation was missing in the yarn.lock. I noticed it because it failed on `--frozen-lockfile`. It's not entirely clear to me at this moment how it was working beforehand (with this entry missing in the lock)

swiftone: Ah.  I'm not bothered by this being 0.3.0 then

packages/configuration-validation/yarn.lock
aarongranick-okta: With updated lock, the integrity has disappeared. Possibly there is a subtle difference between `yarn add` and `yarn install`? Since I had run `yarn add deep-extend -D` at one point, but then removed the dependency from the package.json (replacing with lodash) and ran `yarn install`.

swiftone: What version of yarn are you running?  I believe the integrity check is a newish feature (yarn 1.10)

Issue #547, opened 2019-08-31
Okta-React tokens are not verified and are not revoked.


Itachi-Ichigo:
## I'm submitting this issue for the package(s):

- [ ] jwt-verifier
- [ ] okta-angular
- [ ] oidc-middleware
- [x] okta-react
- [ ] okta-react-native
- [ ] okta-vue

## I'm submitting a:

- [x] Bug report
- [ ] Feature request
- [ ] Other (Describe below)

## Current behavior
Based on my understanding, the okta-token-storage (in local storage for example), is used to store the tokens (Access or/and Id from the authz server). The `isAuthenticated()` function exposed, for any components to check if the User is currently authenticated merely checks the presence of the tokens in local storage and DOES NOT verify them (for instance, by making an introspect call). If I were to manually revoke the tokens and somehow populate the local storage with the previous values or some random values, it will continue to return `isAuthenticated()` = true. 

`isAuthenticated()`: https://github.com/okta/okta-oidc-js/blob/896ca25019621edc04c8576cb6695a6ba7883c6a/packages/okta-react/src/Auth.js#L63

The other issue, is the `logout()` function exposed by OktaAuth, which is meant to sign the user out. But here again, the local storage is cleared and the "session ID" is deleted, but the tokens received as part of login are not really revoked. Basically, the tokens received are still valid after the user logs out. 

`logout()`: https://github.com/okta/okta-oidc-js/blob/896ca25019621edc04c8576cb6695a6ba7883c6a/packages/okta-react/src/Auth.js#L123

## Expected behavior
The expected behavior for `isAuthenticated()` would be to verify the tokens, not just verify their presence.

The expected behavior for 'logout()` would be to revoke the tokens.

## Minimal reproduction of the problem with instructions
`isAuthenticated()`: Using the sample App:
1. login 
2. get into application localstorage using Chrome dev tools
3. edit the okta-token-storage, change the access token value
4. refresh the page, it should return `isAuthenticated()` is true

`logout()`: Using the sample App:
1. login
2. get into app local storage, copy the access token or id token
3. logout
4. manually call your authz server `/introspect` with the copied token, it should show up as active 

## Extra information about the use case/user story you are trying to implement
We are building a react based SPA, server rendered using Express.js. 

## Environment

N/A

Itachi-Ichigo:
I could use the jwt verifier to do the verification manually, but I feel `isAuthenticated` should verify that for me.

aarongranick-okta:
[...]

PR #551, opened 2019-09-09
Handle errors during token renewal. Expose TokenManager [react, angular, vue]

aarongranick-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation changes

Issue Number: 245536

## What is the new behavior?
Exposes the internal `TokenManager` instance via a convenience method for users of the Vue, Angular or React libraries. This allows (among other possibilities), adding event listeners for token expired or login_required events.

- [ ] Yes
- [x] No

## Other information
No new behavior is being added to TokenManager. Additional tests and minor documentation additions related to this task are in PR: https://github.com/okta/okta-auth-js/pull/247


packages/okta-react/src/Auth.js
```diff
@@ -61,9 +75,16 @@ export default class Auth {
   }
 
   async isAuthenticated() {
+    // Support a user-provided method to check authentication
+    if (this._config.isAuthenticated) {
```
robertjd: We should be adding this feature to Vue and Angular as well

packages/okta-react/README.md
```diff
@@ -195,6 +195,10 @@ For PKCE flow, this should be left undefined or set to `['code']`.
     1. `auth.login` is called
     2. SecureRoute is accessed without authentication
 
+- **onTokenError** *(optional)* - callback function. If there is an error while renewing a token, the error will be passed to a handler function. The default handler will call `login()` to initiate a redirect flow. Passing a function here will override the default handler.
```
robertjd: This documentation should be added to Angular and Vue as well

packages/okta-react/src/Auth.js
```diff
@@ -61,9 +75,16 @@ export default class Auth {
...
+      return (this._config.isAuthenticated)();
+    }
+
     // If there could be tokens in the url
     if (location && location.hash && containsAuthTokens.test(location.hash)) return null;
-    return !!(await this.getAccessToken()) || !!(await this.getIdToken());
+
+    // Expect both idToken and accessToken to exist
+    return !!(await this.getAccessToken()) && !!(await this.getIdToken());
```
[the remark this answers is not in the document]
aarongranick-okta: I will revert the logic here so there is no breaking change.

There was a [reported issue](https://github.com/okta/okta-oidc-js/issues/547) about false positives on tokens. Our implementation of `isAuthenticated` is very simple, and can very easily return true when the actual answer (if the jwt verifier is run or the server is asked) is false. 

Ultimately only the server can be the source of truth, and we cannot poll continuously, so there will always be gaps. This line was an attempt to close the gap, but it only handles certain cases, and is still inadequate. 

The main (and possibly only) test comes when the token is renewed. I think orgs which are concerned about false positive states can set the token lifespan to minimum (5 minutes). If they need finer resolution, we could also advise them to poll themselves on `/session/me`. We could also provide some background polling of our own to detect if the user has logged out (or switched identity!)

aarongranick-okta:
@robertjd @swiftone Updated with tests/docs for React, Angular and Vue.

packages/okta-angular/README.md
```diff
@@ -94,6 +94,10 @@ For PKCE flow, this should be left undefined or set to `['code']`.
 - `autoRenew` *(optional)*:
   By default, the library will attempt to renew expired tokens. When an expired token is requested by the library, a renewal request is executed to update the token. If you wish to  to disable auto renewal of tokens, set autoRenew to false.
 
+- `onTokenError` *(optional)* - callback function. If there is an error while renewing a token, the error will be passed to a handler function. The default handler calls `loginRedirect()` to initiate a login flow. Passing a function here will override the default handler.
```
swiftone: Is this an error in renewing (`onRenewError()`?) or an error with the token in general?  If the latter, the documentation does not match that.

packages/okta-react/README.md
```diff
@@ -194,6 +194,10 @@ For PKCE flow, this should be left undefined or set to `['code']`.
     1. `auth.login` is called
     2. SecureRoute is accessed without authentication
 
+- **onTokenError** *(optional)* - callback function. If there is an error while renewing a token, the error will be passed to a handler function. The default handler calls `login()` to initiate a login flow. Passing a function here will override the default handler.
```
swiftone: Same message re: name vs docs

packages/okta-react/test/e2e/harness/src/Home.js
```diff
@@ -19,12 +19,15 @@ export default withAuth(class Home extends Component {
...
     this.state = {
-      authenticated: null
+      authenticated: null,
+      renewMessage: '',
     };
 
     this.checkAuthentication = this.checkAuthentication.bind(this);
     this.login = this.login.bind(this);
     this.logout = this.logout.bind(this);
+    this.renew_idToken = this.renewToken.bind(this, 'idToken');
```
swiftone: Why the underscores?  Mixing snake_case and camelCase leads to platypuses (platypi?)

packages/okta-react/test/jest/auth.test.js
```diff
@@ -345,4 +382,160 @@ describe('Auth component', () => {
...
+        issuer: 'https://foo/oauth2/default',
+        clientId: 'foo',
+        redirectUri: 'https://foo/redirect',
+      });
+      jest.spyOn(auth, 'getAccessToken').mockReturnValue(Promise.resolve(standardAccessTokenParsed));
+      jest.spyOn(auth, 'getIdToken').mockReturnValue(Promise.resolve(null));
+
+      const ret = await auth.isAuthenticated();
+      expect(ret).toBe(true);
+
```
swiftone: nit: why blank line?

packages/okta-react/test/jest/auth.test.js
```diff
@@ -345,4 +382,160 @@ describe('Auth component', () => {
     const authenticated = await auth.isAuthenticated();
     expect(authenticated).toBeFalsy();
   });
+
+  describe('login()', () => {
```
swiftone: I need to save a reference to this test the next time I want a good example of tests in this lib

packages/okta-react/test/support/tokens.js
```diff
@@ -0,0 +1,289 @@
...
+                              'jaWQiOiJQZjBhaWZyaFladTF2MFAxYkZGeiIsInVpZCI6Ij' +
+                              'AwdTFwY2xhNXFZSVJFRExXQ1FWIiwic2NwIjpbIm9wZW5pZ' +
+                              'CIsImVtYWlsIl19.ziKfS8IjSdOdTHCZllTDnLFdE96U9bS' +
+                              'IsJzI0MQ0zlnM2QiiA7nvS54k6Xy78ebnkJvmeMCctjXVKk' +
+                              'JOEhR6vs11qVmIgbwZ4--MqUIRU3WoFEsr0muLl039QrUa1' +
+                              'EQ9-Ua9rPOMaO0pFC6h2lfB_HfzGifXATKsN-wLdxk6cgA';
+
+tokens.standardAccessTokenParsed = {
+  accessToken: tokens.standardAccessToken,
+  expiresAt: 1449703529, // assuming time = 1449699929
```
swiftone: Would it make sense to pull that assumed time into a constant and set everything relative to it, to make the intended time difference more explicitly clear?

packages/okta-angular/README.md
aarongranick-okta: It is an error emitted from the TokenManager. Currently, Renew is the only case where this error will be emitted. Would `onTokenManagerError` be better?

packages/okta-react/test/support/tokens.js
aarongranick-okta: I pulled this file from another repo. We don't actually need real tokens. I am removing it.

aarongranick-okta:
@bretterer updated based on feedback

packages/okta-angular/README.md
swiftone: If it's for all errors and the renew is all that we have now, that's fine as is, I just couldn't tell if we expected future errors from the token manager to come through this callback from the text, so I wanted to be sure.

aarongranick-okta: We could instead have a callback ONLY for the "login_required" error code. (One could still `getTokenManager().on('error', ...` to listen for other errors should they ever exist) This would be almost identical in purpose to the existing callback `onAuthRequired`. In fact we could use that same callback.

aarongranick-okta:
Closed in favor of https://github.com/okta/okta-oidc-js/pull/648

Issue #66, opened 2017-11-29
Add Type Definitions to okta-react so it can be easily be used in a TypeScript project


mraible:
More info: https://www.thepolyglotdeveloper.com/2016/05/add-type-definitions-external-javascript-library-typescript/

tjpeden:
@mraible Angular folks need this too. 👍

nbankin:
Express folks need this too. 👍

robertjd:
Hi all, thanks for the requests!  I certainly love typescript :)  Some questions: I assume you're using VS Code?  If so, I wonder if JsDoc annotations can also serve us well?  I ask because it "feels like" VS Code has been getting better at giving intellisense based on those annotations (but maybe I'm imaging things).

miguelarca:
Any news on getting the Type Definitions yet? I am trying to use it with react and typescript but no luck.

tsiq-swyx:
me too

mraible:
Here's a blog post that shows you how to use okta-react with Create React App and TypeScript: https://developer.okta.com/blog/2017/12/06/bootiful-development-with-spring-boot-and-react

Steps after installing okta-react:

```
yarn add -D @types/react-router-dom@4.2.6
```

You'll get an error when you try to import/use any Okta classes:

```
Could not find a declaration file for module '@okta/okta-react'.
```

Create `src/okta.d.ts` with the following declaration to solve this problem:

```
declare module '@okta/okta-react';
```

Hope this helps!

tsiq-swyx:
wow. surprised its that easy. thanks!

robertjd:
Hi everyone, thank you for your thoughts.  We've discussed this and have decided that Typescript should get some first-party support from us, so we will start working it into our roadmap.  Continue to watch this issue for updates.  Thanks!

royt-via:
hi @robertjd @chrismorris-okta 
do you have any updates about this issue?

swiftone:
@royt-via You can see one of the PRs in the works over at auth-js, which is a dependency of all the front end SDKs (react/vue/angular): https://github.com/okta/okta-auth-js/pull/413

mraible:
TypeScript support was added to Okta React in its v4.0 release. https://github.com/oktadeveloper/schematics/releases/tag/v3.1.0

PR #600, opened 2019-11-12
Add typescript typings for source files

Doff3n:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

No typescript types supported in library

Issue Number: #66

## What is the new behavior?
Typescript typings are generated on build. These types should be further extended
 so that all 'any' declarations has proper types 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/okta-react/package.json
+++ b/packages/okta-react/package.json
@@ -4,8 +4,9 @@
   "description": "React support for Okta",
   "main": "./dist/index.js",
   "scripts": {
-    "build": "rimraf dist/ && yarn build:package-info && babel src -d dist",
+    "build": "rimraf dist/ && yarn build:typescript && yarn build:package-info && babel src -d dist",
     "build:package-info": "node ../../util/write-package-info.js . src/packageInfo.js",
+    "build:typescript" : "npx tsc",
     "lint": "eslint src",
     "lint:watch": "esw -w lib/**",
     "prepare": "yarn build",
@@ -67,6 +68,7 @@
     "react-router-dom": "^4.2.2",
     "rimraf": "^2.6.2",
     "styled-components": "^2.1.2",
+    "typescript": "^3.7.2",
     "webdriver-manager": "^12.1.4"
   },
   "resolutions": {
--- a/packages/okta-react/tsconfig.json
+++ b/packages/okta-react/tsconfig.json
@@ -1,0 +1,18 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "lib": ["es2017", "dom"],
+    "outDir": "./dist",
+    "allowJs": true,
+    "checkJs": false,
+    "strict": false,
+    "noImplicitThis": true,
+    "alwaysStrict": true,
+    "esModuleInterop": true,
+    "noEmit": false,
+    "declaration": true
+  },
+  "include": ["./src/**/*"]
+}
--- a/packages/okta-react/yarn.lock
+++ b/packages/okta-react/yarn.lock
@@ -5048,6 +5048,11 @@ typedarray@^0.0.6:
   version "0.0.6"
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
 
+typescript@^3.7.2:
+  version "3.7.2"
+  resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.7.2.tgz#27e489b95fa5909445e9fef5ee48d81697ad18fb"
+  integrity sha512-ml7V7JfiN2Xwvcer+XAf2csGO1bPBdRbFCkYBczNZggrBZ9c7G3riSUeJmqEU5uOtXNPMhE3n+R4FA/3YOAWOQ==
+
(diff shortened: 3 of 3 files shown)


hboylan:
Looks like there are some failing tests. Maybe related to the TS version bump?

Would be great to have this merged @Doff3n!

edward93:
@hboylan @Doff3n I had this issue before it's most likely because @Doff3n is not a contributer and doesn't have access to OKTA server hence the 401s. If we could get one of the members attention they can migrate this PR under a different user who actually has access to okta server. Adding @swiftone who helped me with my PR and might be able to help us out here too.

danursin:
@Doff3n Thanks for getting this going. Do you have prs upcoming for adding interfaces for component props? I was looking to add them and keep the ts-train moving if you haven't. My impression from you pr was to avoid adding too much in one commit.

Doff3n:
Hi, we have been using Okta for a few months, but we ended up writing our own interfaces and adding them to our code e.g.: 

``` typescript
export interface Auth {
  login(redirectUri: string): void;
  logout(redirectUri: string): void;
  handleAuthentication(): void;
  isAuthenticated(): boolean;
  getAccessToken(): string;
  getUser(): User;
}
```

I think the next step here would be to write typings e.g.: https://github.com/aws/aws-sdk-js/blob/support/typescript/typings/AWS.d.ts, adding typings like the one above. You could start with this PR to generate some typings but it would need to be extended with return types

Or rewrite the code to Typescript starting with the react-okta project for instance and generate typings from that, probably the most clean way of doing this but it requires rewriting and probably rewiring the build.

edward93:
Just like @Doff3n we are using our own types and interfaces, which is not ideal but it gets the job done.
I think migrating `react-okta`'s code base to typescript is ideal solution for us since the project itself is not that large though I'm not sure about tests.

danursin:
Here are the typings I've been using. The HOC types aren't perfect but workable. 
Here is my okta.d.ts
```
declare module "@okta/okta-react" {
    import * as React from "react";
    import { RouteComponentProps } from "react-router-dom";

    export interface OidcConfig {
        issuer: string;
        baseUrl: string;
        redirectUri: string;
        clientId: string;
        pkce: boolean;
        scopes: string | string[];
        disableHttpsCheck?: boolean;
    }

    export interface WithAuthProps {
        auth: Auth;
    }

    export function withAuth<P extends WithAuthProps, C extends React.ComponentType<P>>(
        component: C & React.ComponentType<P>
    ): React.ComponentClass;

    export interface SecurityProps extends Partial<RouteComponentProps>, Partial<OidcConfig> {
        onAuthRequired?: (params: { history: any }) => void;
        auth?: Auth;
    }

    export class Security extends React.Component<SecurityProps, any> {}

    export class SecureRoute extends React.Component<any, any> {}

    export interface ImplicitCallbackProps extends Partial<WithAuthProps> {}

    export class ImplicitCallback extends React.Component<ImplicitCallbackProps, any> {}

    export interface User {
        sub: string;
        name: string;
        locale: string;
        preferred_username: string;
        given_name: string;
        family_name: string;
        zoneinfo: string;
        updated_at: number;
        [key: string]: any;
    }

    export class Auth {
        constructor(config: OidcConfig);

        getTokenManager(): any;
        handleAuthentication(): Promise<void>;
        isAuthenticated(): Promise<boolean | null>;
        getUser(): Promise<User | undefined>;
        getIdToken(): Promise<string | undefined>;
        getAccessToken(): Promise<string | undefined>;
        login(fromUri: string, additionalParams?: any): Promise<void>;
        logout(options?: any): Promise<void>;
        setFromUri(fromUri: string): void;
        getFromUri(): any;
    }
}
```

You can also create a simple hook to consume with the Auth object provided from the `withAuth` HOC with something like this

```
import { useState, useEffect } from "react";
import { Auth, User } from "@okta/okta-react";

interface UseAuthProps {
    isAuthenticated: boolean;
    user?: User;
}

const useAuth = (auth: Auth): UseAuthProps => {
    const [isAuthenticated, setIsAuthenticated] = useState<boolean>(false);

    const [user, setUser] = useState<any>();

    useEffect(() => {
        (async () => {
            const _isAuthenticated = !!(await auth.isAuthenticated());
            if (_isAuthenticated !== isAuthenticated) {
                setIsAuthenticated(_isAuthenticated);
            }
        })();
    });

    useEffect(() => {
        (async () => {
            if (isAuthenticated) {
                const _idToken = await auth.getUser();
                setUser(_idToken);
            } else {
                setUser(undefined);
            }
        })();
    }, [auth, isAuthenticated]);

    return {
        isAuthenticated,
        user
    };
};

export default useAuth;

```

MrMamen:
If you want your own props for the components I think this would be more correct for withAuth:
 ```
 declare module "@okta/okta-react" {
  export function withAuth<P extends WithAuthProps, C extends React.ComponentType<P>>(
    component: C & React.ComponentType<P>
  ): React.ComponentType<Omit<P, keyof WithAuthProps>>;
 ```

The usage would be something like:
``` withAuth<MyComponentProps, typeof MyComponent>(MyComponent)```

ekasprzyk:
Poke on the typescript bindings.  Would be great to have this merged.

PR #669, opened 2020-02-13
[okta-react] Update ImplicitCallback to handle multiple renders

aarongranick-okta:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

If the ImplicitCallback component maps to multiple routes it can render multiple times. However the location can only be read a single time. This can result in <Redirect> being rendered with a null location.

Issue Number: OKTA-277840

## What is the new behavior?
Redirect only occurs if authenticated AND there is a valid location

- [ ] Yes
- [x] No

## Other information
Test harness app has been updated to use the <Switch> component which will prevent multiple render of the ImplicitCallback

## Reviewers
@swiftone

### DIFF
--- a/packages/okta-react/src/ImplicitCallback.js
+++ b/packages/okta-react/src/ImplicitCallback.js
@@ -17,7 +17,6 @@ import withAuth from './withAuth';
 export default withAuth(class ImplicitCallback extends Component {
   constructor(props) {
     super(props);
-
     this.state = {
       authenticated: null,
       error: null
@@ -26,18 +25,23 @@ export default withAuth(class ImplicitCallback extends Component {
 
   componentDidMount() {
     this.props.auth.handleAuthentication()
-    .then(() => this.setState({ authenticated: true }))
+    .then(() => {
+      const location = this.props.auth.getFromUri();
+      this.setState({ authenticated: true, location });
+    })
     .catch(err => this.setState({ authenticated: false, error: err.toString() }));
   }
 
   render() {
-    if (this.state.authenticated === null) {
-      return null;
+    const { error, authenticated, location } = this.state;
+    if (error) {
+      return <p>{error}</p>;
     }
 
-    const location = this.props.auth.getFromUri();
-    return this.state.authenticated ?
-      <Redirect to={location}/> :
-      <p>{this.state.error}</p>;
+    if (authenticated && location) {
+      return <Redirect to={location}/>;
+    }
+
+    return null;
   }
 });
--- a/packages/okta-react/CHANGELOG.md
+++ b/packages/okta-react/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.4.1
+
+### Bug Fixes
+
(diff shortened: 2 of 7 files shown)


swiftone:
Should there be a change to our README/examples to suggest using <Switch>?

aarongranick-okta:
> Should there be a change to our README/examples to suggest using ?

With these changes, it should work OK even with multiple renders. Our documentation/examples usually only show a single route with a callback, so it's not necessarily a problem most users will encounter.

However, I do think we could update README and samples to show "better" practices. I think this would include:
- Using <Switch /> on Routes to avoid multiple render / developer confusion
- Make <Application> component a child of <Router> so that router/history object is available everywhere in the Application and all descendants
- Show `onAuthenticated` callback within a component context (such as member of Application) rather than as a global function

This could be done in another PR, or along with the hooks work you are currently doing.

packages/okta-react/src/ImplicitCallback.js
```diff
@@ -26,18 +25,23 @@ export default withAuth(class ImplicitCallback extends Component {
...
+      this.setState({ authenticated: true, location });
+    })
     .catch(err => this.setState({ authenticated: false, error: err.toString() }));
   }
 
   render() {
-    if (this.state.authenticated === null) {
-      return null;
+    const { error, authenticated, location } = this.state;
+    if (error) {
```
swiftone: is `error` ever temporarily populated?

aarongranick-okta: The logic here should be unchanged from previous behavior. `handleAuthentication()` is called once on `componentDidMount`, if that call throws, the error is put into state where it can be rendered. There is no logic to remove it from state, so it will persist if the component is repeatedly rendered.

swiftone: The order of the logic is different.

Previous:
- If not auth, then null.  Otherwise, if auth succeeded redirect, else show error

Now:
- If error, show error.  Otherwise, if auth and location, redirect.  

An early population of error previously had no effect.  Now it's the first thing checked.

aarongranick-okta: What do you mean by "early population"? `error` and `authenticated` are set at the same time, after the result from `handleAuthentication` I don't see a case where `authenticated` would be null but error would be set. If error is set, I expect `authenticated` to be `false`. Is there another way that `error` can be set, besides the catch block on line 32?

aarongranick-okta:
@swiftone This PR is now green and can be merged. Since it may cause conflicts with work you are doing, I will leave it to you to merge at your convenience

packages/okta-react/src/ImplicitCallback.js
swiftone: That was my question.  If that's not a problem, I'm good.

Issue #656, opened 2020-01-28
`auth.getUser` (from `withAuth`) returns `undefined` if called immediately after sign in


vmaharajh:
## I'm submitting this issue for the package(s):

- [ ] jwt-verifier
- [ ] okta-angular
- [ ] oidc-middleware
- [x] okta-react
- [ ] okta-react-native
- [ ] okta-vue

## I'm submitting a:

- [x] Bug report
- [ ] Feature request
- [ ] Other (Describe below)

## Current behavior
I'm running into some timing issues using the `auth` props provided by `withAuth`.

I use the following to fetch the user and access token from `auth`:

```tsx
    const [user, setUser] = useState<any>();

    useEffect(() => {
      if (props.auth) {
        props.auth.getUser().then(setUser);
      }
    }, [props.auth]);
```

In most cases this works as expected (`setUser` is called with a valid value).

However in the case where my user has just logged in, and is redirected back to my app, `setUser` is called with `undefined`. `auth.getUser` is resolving _successfully_ with an `undefined` value.  In the network tab of F12 dev tools, I don't even see a call to the `/userinfo` endpoint. If I use `setTimeout` to retry in 100ms, the subsequent call succeeds (with the same `auth` object) (and I see a corresponding call to the `/userinfo` endpoint).

## Expected behavior
`auth.getUser` should resolve successfully with a valid user object.

## Minimal reproduction of the problem with instructions
1) Use withAuth to access `auth`, and as soon as it is provided, try to call `getUser`. (A react context provider is a realistic scenario for this (see code below)).
2) Log out the user from the app
3) Log back in, and ensure that the redirect goes to a page that uses the `withAuth` component from step 1)

Here's the code that I'm using for step 1

```tsx
import React from "react";
import { withAuth } from "@okta/okta-react";
import { PropsWithChildren, useState, useEffect } from "react";
import { Auth, AuthContext } from "./Auth";

export interface AuthProviderProps {
  // Injected by withAuth
  auth?: any;
}

/**
 * Injects an `Auth` object in the react context
 */
export const AuthProvider = withAuth(
  (props: PropsWithChildren<AuthProviderProps>) => {
    const [user, setUser] = useState<any>();
    const [accessToken, setAccessToken] = useState<string>();

    const getValue = (): Auth => {
      const result: Auth = {};
      if (user) {
        result.user = { ...user };
      }
      if (accessToken) {
        result.bearerToken = `Bearer ${accessToken}`;
      }
      return result;
[...]

Issue #555, opened 2019-09-15
Support @reach/router with react-router


jackmac92:
Hi,

I'm trying to setup `@okta/okta-react` with a gatsby project. However gatsby uses `@reach/router` to handle routing instead of `react-router`. 

I was looking into forking to support @reach/router, and looking through the source of `@okta/okta-react` it seems [this](https://github.com/okta/okta-oidc-js/blob/32b9a99e3065c34cdcc97018075697356403f11d/packages/okta-react/src/Security.js#L14) use of `withRouter` is the only  aspect that could require some hacking (`@reach/router` does not ship with a comparable HOC). 
But I'm not sure what purpose it currently serves. 

Hoping someone can help me understand what it is doing, I'd be happy to send a PR with any changes I make  to allow support for  `@reach/router`

edit:

to clarify I do know what withRouter  does, just not sure why it is used here

swiftone:
We have an item on our backlog to reduce the dependency on react-router, but it is not yet in progress.  

`withRouter` has the job of wrapping the Auth prop into the route handlers - you don't necessarily need to create an exact parallel if you're working with a router that has different higher-level abstractions.

We're always happy to look at PRs, provided you follow the recommendations in our contribution guidelines: https://github.com/okta/okta-oidc-js/blob/master/CONTRIBUTING.md

jamesliupenn:
@jackmac92 were you able to get it to support `@reach/router`? I'm running into the same issue given the current dependency on `react-router`

aarongranick-okta:
internal ref: OKTA-263509

swiftone:
`@okta/react-router` 2.x no longer requires react-router.

`react-router` is still supported, and is listed as a "peerDependency" (there is no "optionalPeerDependency" that I could find), but version 2.x of SDK should work with other routers (some assembly required).

Issue #499, opened 2019-07-19
okta-react requires react-router (not nextjs et al)


dimacros:
## I'm submitting this issue for the package(s):

- [ ] jwt-verifier
- [ ] okta-angular
- [ ] oidc-middleware
- [x] okta-react
- [ ] okta-react-native
- [ ] okta-vue

## I'm submitting a:

- [ ] Bug report  
- [x] Feature request
- [ ] Other (Describe below)

## Current behavior

It only works with react-router.

## Expected behavior

If we apply the SOLID principles, this library must be independent of the router you use.

## Extra information about the use case/user story you are trying to implement

I think that if the behavior of the library is to authenticate, then it must be agnostic with the library you use to route, but in this case it is not. This library forces me to use react-router and I am currently working with nextjs, you must create a way that is independent to the router or add support for nextjs.

swiftone:
Note: include this in internal discussions about removing dependency on any particular router.

abachuk:
any update on this?

swiftone:
@abachuk - No update other than the JS team agreeing with the request and it is just about getting it to the top of the priority list to fix.  Sorry we don't have better news to report.

swiftone:
Internal ref: 232097

samantha-wong:
+1 on this issue. Would love to be able to use Nextjs with okta.

swiftone:
`@okta/react-router` 2.x no longer requires the use of react-router.

It continues to provide support for react-router, and react-router is listed as a `peerDependency` (there is no "optional peer dependency" option that I'm aware of), but this version should work with other routers without react-router installed.

Issue #429, opened 2019-04-17
Provide hooks api


sibelius:
## I'm submitting this issue for the package(s):

- [ ] jwt-verifier
- [ ] okta-angular
- [ ] oidc-middleware
- [x] okta-react
- [ ] okta-react-native
- [ ] okta-vue

## I'm submitting a:

- [ ] Bug report  
- [x] Feature request
- [ ] Other (Describe below)

## Current behavior

It does not have hooks api

## Expected behavior

It should provide hooks api

`useAuth` for instance

## Environment

- Package Version: 1.2.0
- Browser: chrome
- OS: mac
- Node version (`node -v`): 10.15.0
- Other:

swiftone:
Thanks for the issues and the PR - we're reviewing the involved efforts (including the requirement change for our customers).

swiftone:
`@okta/okta-react` 2.x now includes hooks support along with a number of improvements in the handling of async information.

swiftone:
Thanks to @sibelius for the PR that got it all rolling.

PR #677, opened 2020-02-24
Add: react hooks 

swiftone:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [X] Documentation changes

`okta-react` 1.x requires React Router and offers no support for React Hooks.  

Issue Number: 
 - #429 
- #656 (makes solution more obvious)
- #555 
- #499 
- #405 
- #401 
- #367 

## What is the new behavior?

This makes React Router optional and introduces support for React Hooks while maintaining HOC support for class-based components. 

- [X] Yes
- [ ] No

See "Migration from 1.x to 2.0" in the README

TL;DR: Components specific to react-router will have to be updated to their new versions that are imported in new locations.  Any customizations will have to be updated, though the logic to do so should now be more simple.


swiftone:
Note: Draft PR

- Need to add/update tests
- Need to resolve the `history` that is passed to Auth.js if we aren't using the one that React-Router uses

packages/okta-react/src/withOktaAuth.js
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React from 'react';
+import { useAuth, useAuthState } from './OktaContext';
+
+const withOktaAuth = (ComponentToWrap) => { 
+  // const WrappedComponent = (props) => <ComponentToWrap {...props} auth={auth} authState={authState}/>;
+  return (props) => { 
+    const auth = useAuth();
```
shuowu-okta: `hooks` cannot work in class component, this change may cause the `HOC` fail on class component use case.
https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both

packages/okta-react/src/react-router/index.js
```diff
@@ -10,17 +10,12 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { Component } from 'react';
-import PropTypes from 'prop-types';
+import RouterSecurity from './RouterSecurity';
```
shuowu-okta: Probably can try
```
export { default as RouterSecurity } from './RouterSecurity';
```

packages/okta-react/CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+# 2.0.0
+
+### Features
+
+- Now offers the following React Hooks (2.x requires React 16.8+)
+  - `useAuth`
```
shuowu-okta: How about only provide one hook, then devs can grab what they need from the hook. Something like:
```
const { state, login, ...authfunctions }  = useAuth();
```

packages/okta-react/src/react-router/ImplicitCallback.js
```diff
@@ -0,0 +1,42 @@
...
+import React, { useEffect } from 'react';
+import { Redirect } from 'react-router';
+import { useAuth, useAuthState } from './OktaContext';
+
+const ImplicitCallback = () => { 
+  const auth = useAuth();
+  const authState = useAuthState();
+
+  useEffect( () => {
+    if(authState.isPending) {
```
shuowu-okta: Maybe we should introduce more auth state. Make authentication request during in pending status not make too much sense to me. How about we have `notAuthed`, `authPending`, `authed` (the naming can be revised later)? It can be easier for devs to understand the current auth status, then update UI accordingly.

packages/okta-react/src/Security.js
```diff
@@ -10,34 +10,35 @@
...
-  }
+  console.log('in security render');
 
-  render() {
-    return (
-      <div className={this.props.className}>
-        {this.props.children}
+  useEffect( () => { 
+    let unsub;
+    unsub = auth.on('authStateChange', (newAuthState) => {
```
shuowu-okta: How about leverage the state management practice introduced from react? Current implementation looks like states are coupled with `Auth` class. However, when reading through `Auth` class, seems like it's only a wrapper of auth.js, which probably can be extract out of this react repo in the future.

packages/okta-react/src/Auth.js
```diff
@@ -164,21 +185,44 @@ export default class Auth {
...
     localStorage.removeItem(referrerKey);
     return location;
   }
+
+  on( event, callback ) {
+    // FIXME: restricted events?
+    const eventId = this.eventCount++;
+    this._listeners[event] = this._listeners[event] || {};
+    this._listeners[event][eventId] = callback;
+    return () => {
```
shuowu-okta: If devs eventually start to use this function, return the unregister function may looks like a magic. Provide an `off` function can provide better readability.

packages/okta-react/src/react-router/index.js
swiftone: We could, but what would be the benefit?  They get to type more?

packages/okta-react/CHANGELOG.md
swiftone: The main reason I separated them is that they serve different purposes.  `useAuth` allows you to manipulate the auth state, while `useAuthState` is for READING the auth state.  That means the difference serves a communication purpose - someone reviewing a component that manipulates the auth state has that jump out as obvious.

packages/okta-react/src/react-router/ImplicitCallback.js
swiftone: How are you seeing the difference between "pending" and "notAuthed" here?  To the dev, the result is the same - their content needs to wait.

packages/okta-react/src/Security.js
swiftone: Can you clarify what that would look like?  
- We need it to trigger on the initial render
- We need it to update after an async process
- We need to make sure later changes in state DON'T trigger a recheck of the auth

packages/okta-react/src/Auth.js
swiftone: This is a fairly common pattern, and it shelters us from exposing the internal `eventId` (which means we can later change the eventId mechanism).  

I'll do a little research into alternatives.

packages/okta-react/src/react-router/ImplicitCallback.js
shuowu-okta: I see `pending` as in progress, which means a spinner probably will be added in UI. 
`notAuthed` mostly means auth not start yet, or auth failed, which is a result status.

packages/okta-react/src/react-router/index.js
shuowu-okta: not really, it can end up with only 3 lines in the index file.

packages/okta-react/src/Auth.js
shuowu-okta: I understand this pattern definitely work. Also if it's only private method for SDK internally, how about add _ to mark it as private.

packages/okta-react/src/Security.js
shuowu-okta: My idea is we can leverage `context` and `useReducer` to store/update state globally. I learned the pattern from https://medium.com/simply/state-management-with-react-hooks-and-context-api-at-10-lines-of-code-baf6be8302c
It works like `redux`, but without `redux` dependency.

packages/okta-react/src/Auth.js
swiftone: The method is public, though as a return value it is without a name, so I'm not sure I follow.

If you mean making the eventId intoi `_eventId`, that's a good call.

packages/okta-react/src/Security.js
swiftone: I'm still not quite following.  That article uses `useState`, `useReducer` and a context.  

I'm using `useState`, `useEffect`, and a context.  

Even if I used `useReducer`, I'd still have to subscribe to the event emitter, so I'd still have to have `useEffect`.  It looks like what you're suggesting simply adds a layer over being able to update state when the status changes.  What am I missing?

packages/okta-react/src/react-router/ImplicitCallback.js
swiftone: > not start yet, or auth failed

There are 4 potential states:

- not yet checked for auth - this is when the SPA code has just loaded, the user may or may not ALREADY be authenticated (they may have just hit "reload" for example, or been redirected to the app as a full page load from another of their own pages)
- Waiting for the async processes of getting a token from the token manager (this starts immediately after the above)
- checked for auth and it succeeded
- checked for auth and it failed

This code treats the first two as `isPending` because the user literally takes no action to move from one to the other and the visuals should not be different in either case. 

What would a dev do differently if the two `isPending` cases were broken apart?

packages/okta-react/CHANGELOG.md
shuowu-okta: I get your point. But still feel it might be easier just provide one single hook, since state can still be accessed very easily by destructing the return of hook.

packages/okta-react/src/Auth.js
shuowu-okta: My question was about the access modifier of `on` method.

packages/okta-react/src/Security.js
shuowu-okta: My suggestion is put `useReducer` in `Provider`, then you can have access to the global state (authState) and the `dispatch` function to make state change from context. Then by following the `flux` flow (action -> reducer) should be able to make state change properly. I am not against the `observer pattern` introduced in this PR, just because this pr is targeting to specific framework, it might be better to follow the popular pattern in the framework's ecosystem.

packages/okta-react/src/react-router/ImplicitCallback.js
shuowu-okta: Thanks for the explanation, the states make sense to me now. 
For the logic here in code, does it mean it still can make async request for token process even it's already in state 2?

packages/okta-react/CHANGELOG.md
swiftone: "easier" in the sense of having one object for two purposes, sure.  But as I pointed out, that comes at a cost of communication, which is the goal of good programming - to communicate.

packages/okta-react/src/Security.js
swiftone: My concern is that we'd be adding an abstraction layer to cover just one case - where we use the dispatcher in a callback to this event emitter.

Not having a reducer at all and just updating state is also a popular pattern, used when you don't have a complex state that involves multiple actions.  

Why do you think this state management is complex enough to justify a reducer?

packages/okta-react/src/Auth.js
swiftone: `on()` is a public method, albeit one that would be used rarely, since most users should be fine with the provided `<Security>` component.

packages/okta-react/src/react-router/ImplicitCallback.js
swiftone: It could, but we're using `useEffect()` so that it won't. Even if we had the first two states as different states, they could still trigger multiple async checks because the change passing through re-renders means you can't guarantee the order of events. I'm shooting for simplicity and what the dev needs to know: Has authentication been figured out or not.

packages/okta-react/src/react-router/index.js
swiftone: I'm not seeing the benefit of adopting an unusual syntax that doesn't add clarity for the sake of a few lines.

shuowu-okta: It's just a FYI comment. It has been adopted by some popular open source repos.

swiftone: oh?  Any examples?  I'm stumped as to why this syntax would be more popular, so examples can help.

packages/okta-react/src/Security.js
shuowu-okta: By browsing through all three packages for `react`, 'angular' and `vue`. My feeling is we probably can extract the `Auth.js` out from framework repos, since they are all doing basically same work. If that can be the case in the future, I would suggest not couple state management with `Auth.js`.

packages/okta-react/src/react-router/index.js
shuowu-okta: I learned it from `material-ui` repo. Here is an example https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/AppBar/index.js

packages/okta-react/src/Security.js
swiftone: Careful: the devil is in the details.  Auth.js translates between authJS and the application, and angular, vue, and react have some different expectations.

The current React SDK has no state management in AuthJS, and it is a nightmare of async calls that is a hotbed of race conditions.

Without being too abstract, what are the exact lines you see separating a unified Auth.js from the three SDKs?  (and yet, separating it from the truly generic authjs?)

packages/okta-react/src/react-router/ImplicitCallback.js
shuowu-okta: I still think it should be able to reduce most unnecessary async calls with one more state provided globally. 
If it's for simplicity concern, I am ok with that.

packages/okta-react/CHANGELOG.md
shuowu-okta: Agreed. As we are rolling out new contract for dev, just want to explore other options a bit.

packages/okta-react/src/Security.js
shuowu-okta: Agree, over abstract across repos without careful design can be a maintenance nightmare. I may revisit this idea in the future.

packages/okta-react/package.json
```diff
@@ -38,8 +38,8 @@
     "prop-types": "^15.5.10"
   },
   "peerDependencies": {
-    "react": ">=15",
-    "react-router": ">=4"
+    "react": ">=16.8.0",
+    "react-router-dom": ">=5.1.0"
```
shuowu-okta: Should we handle console warning if other router been selected by dev? Also if I remember correctly `npm2` will get peerDependencies installed automatically, do we need to handle this scenario?

swiftone: I don't believe we support npm2 - that was quite a while ago.

I did some research, and there doesn't appear to be an option.  `optional` modules are installed if available, and `peerDependencies` are warnings.  The common approach seems to be to put in it `peerDependencies` and document.

packages/okta-react/CHANGELOG.md
aarongranick-okta: Maybe `useAuthService` would communicate more clearly? (I sometimes get confused referring to `Auth.js` (with a capital 'A') because of ambiguity with the underlying okta-auth-js. We could rename it to `AuthService`, this would match the naming in angular and indicates it provides methods (which can produce side effects. reading authState does not change anything).

swiftone: There's a lot there, let's break it down:

### rename Auth to AuthService

I'm (mostly) fine with this.  The only reason I'm mostly fine is I don't know that the "AuthService" concept is generally understood - I tend to the think of services in terms of web calls and not generic functionality providers, but I don't know if that's just me.

If we do this, it should be consistent throughout:
- Auth.js to AuthService.js
- `auth` param to `<Security>` becomes `authService`
- `useAuth` becomes `useAuthService`
- prop becomes `authService` instead of `auth`
- I'd leave `withOktaAuth` the same and still provide both props (though change `auth` to `authService` unless convinced otherwise (see below)

Does this make sense?  If not, I'd say leave it as-is.    

### The HOC

I have one HOC but two Hooks because the HOC is kind of annoying to use more than one of. ( `annoying(toUseMoreThen(OneOf)) `  

I think there's a communication benefit to having two hooks - using just `authState` says you aren't changing the state, just reading it, while `auth` (or `authService`) declares you are changing it, making it obvious to anyone skimming the code.    (userinfo is a hole here, but we've talked about putting it on the authState in the future)

I couldn't tell from your comment which way your were advocating.  Given my arguments, do you have a stance one way or the other?

packages/okta-react/src/Security.js
aarongranick-okta: @swiftone  & I did have some discussions about having the state separated from the service, following a "store/emitter" pattern for the state. The article @shuowu-okta  linked to has the same idea, but adds the "reducer" concept (which I think is probably not necessary) and uses the context as the store (which is a minor React-specific optimization). We *could* put the current logic into a barebones "AuthStateProvider" or "AuthStateStore" which holds the state, leaving the Auth wrapper stateless.

I think that ultimately we would like to normalize the services and push the implementation down a layer to `okta-auth-js` so that these SDK's can be simplified. Our users could refer to a single set of documentation for configuration and API methods in `okta-auth-js`, and the individual SDK's would not be hiding anything via a containing wrapper.

Incidentally, the underlying service [DOES have an emitter](https://github.com/okta/okta-auth-js/blob/master/packages/okta-auth-js/lib/browser/browser.js#L160), used by the TokenManager. I was *this* close to adding an `on` method on the core service, but I didn't have a specific event to expose. I actually think this `authStateChange` event  is a GREAT candidate for something we can push down into the core service. It's a VERY useful addition. Why should it only be for React? 

After this release we could take this functionality and plug it in directly to `okta-auth-js`, it would not change the contract we are putting forward here. You would still use `auth.on(`authStateChange`...`

packages/okta-react/CHANGELOG.md
aarongranick-okta: I like everything about changing "auth" to "authService". We are using "Auth Service" in the README because "Auth Object" doesn't mean much. I understand a service to be a singleton object which exposes public methods. It is something you *use*.

I understand what you are saying about communication, but I think a single effect will be easy to understand and we would future-proof ourselves better with a single effect that passes ALL the objects that we are providing now or in the future. Our docs should show using the array syntax `['auth', 'authState']` on your effect so you only re-render on the props that you care about. I think this would be a good practice vs. rendering whenever "anything" changes. I also like the symmetry between the effect and the HOC. If we add another property, we will add it to both.

packages/okta-react/src/react-router/ImplicitCallback.js
aarongranick-okta: Currently all the logic to update the state is in the method `updateAuthState`. If I understand correctly, @shuowu-okta is asking for an "initial" state, before this method has been called. Then when this method is called it would immediately emit a "pending" state before beginning any async logic. At the end of the logic, there will be another update and you will either be authenticated or not.  One advantage of this is that `updateAuthState` could check the current state, if it is "pending" that means there is another `updateAuthState` running and it could exit immediately. We avoid potentially having two async processes running at the same time, which I think could easily happen if the method is being called from a component.

packages/okta-react/CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+  - React Router 5 continues to be supported, but is now optional
+  - Routers other than React-Router will have to write their own version of `LoginCallback` component 
+
+### Breaking Changes
+- See the `Migration from 1.x to 2.0` section of the README for details on migrating your applications
+  - `Auth.js` and the `auth` parameter to `<Security>` have been renamed to `AuthService.js` and `authService`
+  - `<ImplicitCallback>` has been replaced with `<LoginCallback>`
+  - `auth.IsAuthenticated()` has been removed
+  - `withAuth` has been replaced with `withOktaAuth`, which gives slightly different parameters
+  - the arguments passed to the optional `isAuthenticated()` callback provided to the `<Security>` component have changed
```
aarongranick-okta: I think you mean `onAuthRequired()` callback?

packages/okta-react/README.md
```diff
@@ -55,26 +63,49 @@ npm install --save @okta/okta-react
...
-- [SecureRoute](#secureroute) - A normal `Route` except authentication is needed to render the component.
-- [ImplicitCallback](#implicitcallback) - A simple component which handles the login callback.
+`okta-react` provides the necessary tools to build an integration with most common React-based SPA routers.
 
-## Create Routes
+- [Security](#security) - Allows you to supply your OpenID Connect client [configuration](#reference). Includes React context providers to allow the use of the [useOktaAuth][] React Hook, or the [withOktaAuth][] Higher Order Component wrapper.
+- [LoginCallback](#LoginCallback) - A simple component which handles the login callback when the user is redirected back to the application from the Okta login site.
 
-Here is a minimal working example. This example defines 3 routes:
+Users of routers other than `react-router` will have to add their own checks for authentication.  The exact method will depend on your router, but the basic idea is to use [useOktaAuth][] to see if a `authState.isPending` is false and `authState.isAuthenticated` is true.  If both are false, you can send them to login via `authService.login(...)`.
```
aarongranick-okta: Instead of "will have to add...", how about something like:
"can use `authState` to check authentication before loading a route"

packages/okta-react/README.md
```diff
@@ -55,26 +63,49 @@ npm install --save @okta/okta-react
...
+
+### Available Hooks
+
+These hooks require that the component be inside a `Security` component that provides the necessary context.  Class-based components can gain access to the same information via the `withOktaAuth` Higher Order Component, which provides `authService` and `authState` as props to the wrapped component.
```
aarongranick-okta: Instead of "be inside a" how about "are a child or descendant of"

packages/okta-react/README.md
```diff
@@ -55,26 +63,49 @@ npm install --save @okta/okta-react
...
+
+- [useOktaAuth](#useOktaAuth) - gives an object with two properties:
+  - `authService` - the [Auth service][] object that allows for modification of and deeper queries into the authentication state
```
aarongranick-okta: The Auth service does a lot of stuff, maybe we can keep it very brief here and just link to the section where we describe it in detail

packages/okta-react/README.md
```diff
@@ -97,69 +128,106 @@ class App extends Component {
...
+```jsx
+import React from 'react';
+import { SecureRoute } from '@okta/okta-react/react-router';
+import { Security, LoginCallback } from '@okta/okta-react';
+import Home from './Home';
+import Protected from './Protected';
 
+const App = () => (
+  <Router>
+    <RouterSecurity
```
aarongranick-okta: Should be `<Security>`?

packages/okta-react/README.md
```diff
@@ -97,69 +128,106 @@ class App extends Component {
...
-
-  async componentDidUpdate() {
-    this.checkAuthentication();
+    this.props.authService.logout('/');
   }
 
   render() {
-    if (this.state.authenticated === null) return null;
-    return this.state.authenticated ?
+    if (this.props.authState.isPending) return <div>Loading...</div>;
```
aarongranick-okta: This whole example is MUCH improved by these authState changes! I hated that `checkAuthentication` on `componentDidUpdate`! Our users are going to love this.

packages/okta-react/README.md
```diff
@@ -182,20 +250,55 @@ export default withAuth(class MessageList extends Component {
...
+  );
+  return <ul>{items}</ul>;
+};
+```
 ## Reference
 
 ### `Security`
 
-Security is the top-most component of okta-react. This is where most of the configuration is provided.
+`<Security>` is the top-most component of okta-react. This is where most of the configuration is provided.  Apps using `react-router` should use `RouterSecurity` from `@okta/okta-react/react-router` as drop-in-replacement for `<Security>`
```
aarongranick-okta: This bit about `<RouterSecurity>` can be removed

packages/okta-react/README.md
```diff
@@ -301,7 +399,7 @@ export default App;
 
 #### PKCE Example
 
-Assuming you have configured your application to allow the `Authorization code` grant type, simply pass `pkce=true` to the `Security` component. This will configure the [Auth service][] to perform PKCE flow for both login and token refresh.
+Assuming you have configured your application to allow the `Authorization code` grant type, simply pass `pkce=true` to the `Security` component. This will configure the [Auth service][] to perform PKCE flow for both login and token refresh.
```
aarongranick-okta: This is fine, I don't actually see any changes here, so no need to change it in this PR. but I think PKCE is now our default and implicit is effective deprecated. This is what the admin UI looks like now. 
Screen Shot 2020-02-28 at 12 39 10 PM

packages/okta-react/README.md
```diff
@@ -352,76 +450,165 @@ class App extends Component {
 
 `SecureRoute` ensures that a route is only rendered if the user is authenticated. If the user is not authenticated, it calls `onAuthRequired` if it exists, otherwise, it redirects to Okta.
 
-### `ImplicitCallback`
+`SecureRoute` is specific to `react-router`.  Other routers will need their own methods to ensure authentication using the hooks/HOC props provided by this SDK.
```
aarongranick-okta: How about "was built specifically to integrate with `react-router`. If you are using a different router, you will need to..."

packages/okta-react/README.md
```diff
@@ -352,76 +450,165 @@ class App extends Component {
...
+
+### `LoginCallback`
 
-`ImplicitCallback` handles the callback after the redirect. By default, it parses the tokens from the uri, stores them, then redirects to `/`. If a `SecureRoute` caused the redirect, then the callback redirects to the secured route. For more advanced cases, this component can be copied to your own source tree and modified as needed.
+`LoginCallback` handles the callback after the redirect to and back from the Okta site. By default, it parses the tokens from the uri, stores them, then redirects to `/`. If a `SecureRoute` caused the redirect, then the callback redirects to the secured route. For more advanced cases, this component can be copied to your own source tree and modified as needed.
```
aarongranick-okta: "from the Okta-hosted login page"?

packages/okta-react/README.md
```diff
@@ -352,76 +450,165 @@ class App extends Component {
...
 
-### `withAuth`
+### `withOktaAuth`
 
-`withAuth` is a [higher-order component][] which injects an `auth` prop into the component. This provides a way for components to make decisions based on auth state.
+`withOktaAuth` is a [higher-order component][] which injects an `authService` prop and an `authState` prop into the component. Function-based components will want to use the `useOktaAuth` hook instead.  These props provide a way for components to make decisions based on auth state or to call [Auth Service][] methods, such as `.login()` or `.logout()`.  Components wrapped in `withOktaAuth()` need to exist inside a `<Security>` component to have the necessary context.
```
aarongranick-okta: "inside a <Security> component" -> "must be a child or descendant of a <Security> component"

packages/okta-react/README.md
```diff
@@ -352,76 +450,165 @@ class App extends Component {
...
 ```
 
 > Note: For information on obtaining a `sessionToken` using the [Okta Sign-In Widget](https://github.com/okta/okta-signin-widget), please see the [`renderEl()` example](https://github.com/okta/okta-signin-widget#rendereloptions-success-error).
 
-#### `auth.handleAuthentication()`
+#### `authService.handleAuthentication()`
+
+Parses tokens from the url and stores them.  Done when handling a redirect back to the application from the issuer.  See the LoginCallback component for an example of use.
+
+### `authService.setFromUri(uri, queryParams)`
```
aarongranick-okta: This method does not take `queryParams` anymore. It takes a URI string. If a relative path is passed, it will be converted to an absolute URI before it is stored.

packages/okta-react/README.md
```diff
@@ -352,76 +450,165 @@ class App extends Component {
...
 
-Parses tokens from the url and stores them.
+Store the current URL state before a redirect occurs.  Called internally by `auth.login`.
+
+#### `authService.getFromUri()`
+
+Returns the stored URI and query parameters stored by `setFromUri` and removes it from storage.  See the `LoginCallback` component for an example of use.
```
aarongranick-okta: "returns the stored URI **string**"

packages/okta-react/CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+  - error handling for authentication is now handled by putting the error into the `authState.error` property
+
```
aarongranick-okta: `setFromURI` now takes a string (instead of object). `getFromURI` returns a string (instead of object)

packages/okta-react/README.md
swiftone: > `SecureRoute` integrates with `react-router`.  Other routers will need their own methods to ensure authentication using the hooks/HOC props provided by this SDK.

packages/okta-react/src/AuthService.js
```diff
@@ -94,32 +107,55 @@ class AuthService {
 
   emitAuthState(state) { 
     this._authState = state;
-    this.emit('authStateChange', this._authState);
+    this.emit('authStateChange', this.getAuthState());
```
shuowu-okta: Should we scope the emitted event with `okta`?

swiftone: As this event will be emitted on `authService`, which you got from either `useOktaAuth` or `withOktaAuth`, this falls in the pattern of having Okta at the highest level but not redundantly at every level.

packages/okta-react/src/AuthService.js
```diff
@@ -43,45 +41,123 @@ export default class Auth {
...
+    } catch (error) { 
+      this._pending.authStateUpdate = null;
+      this.emitAuthState({ 
+        isAuthenticated: false,
+        error,
+        idToken: null,
+        accessToken: null,
+      });
+    } 
+    authStateUpdate.resolve();
```
aarongranick-okta: I think `resolve()` needs to be called with a value, otherwise the promise will resolve `undefined`. This will break `handleAuthentication` which is expecting a value to be returned: https://github.com/okta/okta-oidc-js/pull/677/files#diff-8a51cbb87b5cfd7e092f0410797f00e9R87
`handleAuthentication` should either read from `authState.getAuthState()` (and this promise is defined as always returning undefined), or the promise should resolve to the authState.

swiftone: Ah, I missed a place to update.

I was removing all the numerous places authState was returned to make it clear to consult the one source of truth (authService.getAuthState()), but I missed this update.

Once more unto the breach.

packages/okta-react/README.md
```diff
@@ -351,77 +446,175 @@ class App extends Component {
...
+
+#### `authService.getTokenManager()`
+
+(async method) Resolves with the internal [TokenManager](https://github.com/okta/okta-auth-js#tokenmanager).
 
-Parses tokens from the url and stores them.
+#### `authService.updateAuthState()`
 
-#### `auth.setFromUri(uri, queryParams)`
+(async method) Triggers a re-evaluation of whether a user is considered authenticated.  Might be need to be triggered manually for users that pass overrides to `isAuthenticated` to the Security component or to the `Auth` constructor.  Does NOT perform a login, simply re-evaluates the current authenticated status. An 'authStateChange' event is emitted once the re-evaluation is complete.
```
aarongranick-okta: I think again it is not `isAuthenticated`, but `onAuthRequired` where you might want to call this function. It is called  by us right before we call `onAuthRequired`, using either the value passed from the SecureRoute or window.location.origin if no value was provided. A developer could make their own call to this method to cause a redirect to a different path after login. https://github.com/okta/okta-oidc-js/pull/677/files#diff-8a51cbb87b5cfd7e092f0410797f00e9R205

packages/okta-react/README.md
```diff
@@ -351,77 +446,175 @@ class App extends Component {
...
+(async method) Resolves with the internal [TokenManager](https://github.com/okta/okta-auth-js#tokenmanager).
 
-Parses tokens from the url and stores them.
+#### `authService.updateAuthState()`
 
-#### `auth.setFromUri(uri, queryParams)`
+(async method) Triggers a re-evaluation of whether a user is considered authenticated.  Might be need to be triggered manually for users that pass overrides to `isAuthenticated` to the Security component or to the `Auth` constructor.  Does NOT perform a login, simply re-evaluates the current authenticated status. An 'authStateChange' event is emitted once the re-evaluation is complete. 
 
-Store the current URL state before a redirect occurs.
+To get the updated auth status you must call [authService.getAuthState()](#authservicegetauthstate) AFTER the `authStateChange` event has been emitted.  For straight-forward React applications using the authState in the front end, this is done for you in the `<Security>` component and the `authState` reported by the [useOktaAuth()](#useoktaauth) React Hook or [withOktaAuth()](#useoktaauth) HOC will automatically refresh to the latest [authState](#authstate).
```
aarongranick-okta: Does this paragraph belong in the section below (authService.on) ? (Actually it seems redundant with the text below, which sums it all up pretty nicely)

packages/okta-react/README.md
```diff
@@ -351,77 +446,175 @@ class App extends Component {
...
+
+Any component that was using `withAuth()` to get the `auth` object and called the properties above has two options to migrate to the new SDK:
+1. Replace the use of `withAuth()` with [withOktaAuth()](#withoktaauth), and replace any of these asynchronous calls to the `auth` methods with the values of the related [authState](#authstate) properties. 
+**OR**
+2. Remove the use of `withAuth()` and instead use the [useOktaAuth()](#useoktaauth) React Hook to get the [authService](#authservice) and [authState](#authstate) objects.  Any use of the `auth` methods (`.isAuthenticated()`, `.getAccessToken()`, and `.getIdToken()`) should change to use the already calculated properties of [authState](#authstate).
+
+Either of these options needs to be a descendant of a `<Security>` component to have the necessary context.
```
aarongranick-okta: "Using either of these options, **your component** needs to be a descendant of..."

Issue #668, opened 2020-02-13
Deprecated packages


ashleym1972:
## I'm submitting this issue for the package(s):

- [x] jwt-verifier
- [ ] okta-angular
- [ ] oidc-middleware
- [ ] okta-react
- [ ] okta-react-native
- [ ] okta-vue

## I'm submitting a:

- [x] Bug report  
- [ ] Feature request
- [ ] Other (Describe below)

## Current behavior
```
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated popper.js@1.16.1: You can find the new Popper v2 at @popperjs/core, this package is dedicated to the legacy v1
npm WARN deprecated core-js@2.6.11: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
```

## Expected behavior
No deprecated packages

## Environment

- Package Version: 1.0.0
- Browser:
- OS:
- Node version (`node -v`):
- Other:

bretterer:
@ashleym1972 - Thank you for your report. We will take a look at our jwt-verifier to make sure that we get rid of the deprecated packages. 

Do these deprecation messages block you in any way with your project?

internal ref: OKTA-278077

aarongranick-okta:
Closing issue as deprecated packages have been upgraded/removed.

PR #678, opened 2020-02-24
test(jwt-verifier): resolve deprecated packages - OKTA-278077

shuowu-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: #668 

## What is the new behavior?
`node-fetch` is used to replace deprecated `request` module.

- [ ] Yes
- [x] No

## Reviewers
@aarongranick-okta 
@swiftone

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -96,5 +96,28 @@
       },
       "disableOptimisticBPs": true
     }, 
+    {
+      "type": "node",
+      "request": "launch",
+      "protocol": "auto",
+      "name": "jwt-verifier Jest",
+      "cwd": "${workspaceFolder}/packages/jwt-verifier",
+      "args": [
+        "node_modules/.bin/jest",
+        "--runInBand",
+        "--no-cache",
+        "${workspaceFolder}/${relativeFile}"
+      ],
+      "sourceMaps": true,
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "env": {
+
+      },
+      "sourceMapPathOverrides": {
+        "webpack:///*": "/*",
+      },
+      "disableOptimisticBPs": true
+    }, 
   ]
 }
--- a/packages/jwt-verifier/README.md
+++ b/packages/jwt-verifier/README.md
@@ -131,6 +131,18 @@ const oktaJwtVerifier = new OktaJwtVerifier({
 });
 ```
 
+## Testing
+Setup SPA and Web App in your Okta org and testing environment variables by following [Testing](https://github.com/okta/okta-oidc-js#testing) section in Monorepo's README.
+
+**NOTE:** 
+
+When create SPA application in your Okta org, please make sure all `Implicit` checks have been checked in `General Settings -> Application -> Allowed grant types` section.
+
+Command for running unit test:
+```
+yarn test:unit
+```
+
 ## Contributing
 We welcome contributions to all of our open-source packages. Please see the [contribution guide](https://github.com/okta/okta-oidc-js/blob/master/CONTRIBUTING.md) to understand how to structure a contribution.
 
--- a/packages/jwt-verifier/README.md
+++ b/packages/jwt-verifier/README.md
@@ -132,7 +132,7 @@ const oktaJwtVerifier = new OktaJwtVerifier({
 ```
 
 ## Testing
-Setup SPA and Web App in your Okta org and testing environment variables by following [Testing](https://github.com/okta/okta-oidc-js#testing) section in Monorepo's README.
+Setup SPA and Web App in your Okta org and testing environment variables by following [Testing](https://github.com/okta/okta-oidc-js#testing) section in okta-oidc-js Monorepo's README.
 
 **NOTE:** 
 
--- a/packages/jwt-verifier/test/util.js
+++ b/packages/jwt-verifier/test/util.js
@@ -42,6 +42,10 @@ function getAccessToken(options = {}) {
 
       return resp.json();
     }).then(body => {
+      if (!body.sessionToken) {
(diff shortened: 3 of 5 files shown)


packages/jwt-verifier/README.md
```diff
@@ -131,6 +131,18 @@ const oktaJwtVerifier = new OktaJwtVerifier({
 });
 ```
 
+## Testing
+Setup SPA and Web App in your Okta org and testing environment variables by following [Testing](https://github.com/okta/okta-oidc-js#testing) section in Monorepo's README.
+
+**NOTE:** 
+
+When create SPA application in your Okta org, please make sure all `Implicit` checks have been checked in `General Settings -> Application -> Allowed grant types` section.
```
swiftone: Is this an actual requirement?  If someone is only requesting `access token`, is `id token` grant required?

packages/jwt-verifier/test/util.js
```diff
@@ -49,25 +52,28 @@ function getAccessToken(options = {}) {
...
-        const match = resp.headers.location.match(/access_token=([^&]+)/);
-        const accessToken = match && match[1];
-        if (!accessToken){
-          return reject(new Error('Could not parse access token from URI'));
-        }
-        resolve(accessToken);
-      })
+      return fetch(authorizeUrl, { redirect: 'manual' });
+    }).then(resp => {
+      if (!resp.status >= 400) {
```
swiftone: Why not use `resp.ok`?

packages/jwt-verifier/README.md
```diff
@@ -131,6 +131,18 @@ const oktaJwtVerifier = new OktaJwtVerifier({
 });
 ```
 
+## Testing
+Setup SPA and Web App in your Okta org and testing environment variables by following [Testing](https://github.com/okta/okta-oidc-js#testing) section in Monorepo's README.
```
swiftone: Would it be more clear to say 'okta-oidc-js Monorepo's README'?

packages/jwt-verifier/test/util.js
shuowu-okta: `resp.ok` only serve for range `200-299`, since we need handle redirect logic in the fetch, I think 400 can be used to exclude redirect status code.

packages/jwt-verifier/README.md
shuowu-okta: I added `note` here only because it prevent me from running tests successfully without checking `Implicit `. I can remove it if it's not the proper way to declare it. Any suggestions?

swiftone: I'd move that to the `Contributing` section and/or make it clear that they need to be enabled for the tests.  General devs won't be running our tests.

Otherwise this looks like they can't use jwt-verifier without these flags.

packages/jwt-verifier/test/util.js
```diff
@@ -27,18 +27,21 @@ function getAccessToken(options = {}) {
...
       }
-    }, function (err, resp, body) {
-      const statusCode = resp ? resp.statusCode : null;
-      if (err || (resp && resp.statusCode >= 400) || !resp) {
-        console.error('/api/v1/authn returned error: ', statusCode, err || body);
-        return reject(new Error('/api/v1/authn error: ' + (err || body)));
-      }
+
+      return resp.json();
+    }).then(body => {
```
aarongranick-okta: Possibly worth considering. Some errors will be returned with a 200 response code. In this case there will be no sessionToken and there should be an error code / error summary.  See how SIW handles this: https://github.com/okta/okta-signin-widget/blob/master/src/util/Util.js#L79  We might want to avoid making the next request if there is no sessionToken, and instead throw an error. 

It is also possible (though less likely) that the server will return a 200 status code but invalid JSON, in which case `resp.json()` will throw. This case I think we can leave unhandled because it will already throw. But we are handling it in AuthJS 3.0 (for the sake of better errors in the SIW): https://github.com/okta/okta-auth-js/pull/326/files#diff-ed946c93045cf8b90899ada2e4655327R21

.vscode/launch.json
```diff
@@ -90,6 +90,29 @@
...
+        "node_modules/.bin/jest",
+        "--runInBand",
+        "--no-cache",
+        "${workspaceFolder}/${relativeFile}"
+      ],
+      "sourceMaps": true,
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "env": {
+
```
aarongranick-okta: Thanks for adding this!

packages/jwt-verifier/test/util.js
aarongranick-okta: Should this be `(resp.status >= 400)` ? Without the `!` ?

shuowu-okta: Good catch! Thanks

packages/jwt-verifier/test/util.js
shuowu-okta: Good point. As the change is only for a util function in test, i would propose to handle empty `sessionToken`, then leave invalid json to `catch`.

aarongranick-okta: Yeah it would be only to help us with better error message in the test logs.

packages/jwt-verifier/README.md
shuowu-okta: As it's under `tests` section, just want to confirm is it fine to keep it unchanged?

packages/jwt-verifier/README.md
shuowu-okta: Updated

PR #680, opened 2020-02-26
refactor: remove Router dep from Security, Auth, ImplicitCallback

aarongranick-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [] Other... Please describe:

Currently there is a dependency on react-router within `Security`, 'Auth` and `ImplicitCallback`
Issue Number: N/A

## What is the new behavior?
Only `SecureRoute` has a dependency on router

- [x] Yes
- [ ] No

The signature for `onAuthRequired` has changed and the `history` object is no longer passed to the callback function.

`getFromUri()` and `setFromUri()` now set and return the URI as a string (rather than an object)

## Reviewers
@swiftone 
@shuowu-okta

### DIFF
--- a/packages/okta-react/README.md
+++ b/packages/okta-react/README.md
@@ -73,7 +73,7 @@ Here is a minimal working example. This example defines 3 routes:
 // src/App.js
 
 import React, { Component } from 'react';
-import { BrowserRouter as Router, Route } from 'react-router-dom';
+import { Route, withRouter } from 'react-router';
 import { Security, SecureRoute, ImplicitCallback } from '@okta/okta-react';
 import Home from './Home';
 import Protected from './Protected';
@@ -81,7 +81,6 @@ import Protected from './Protected';
 class App extends Component {
   render() {
     return (
-      <Router>
         <Security issuer='https://{yourOktaDomain}.com/oauth2/default'
                   clientId='{clientId}'
                   redirectUri={window.location.origin + '/implicit/callback'} >
@@ -89,12 +88,23 @@ class App extends Component {
           <SecureRoute path='/protected' component={Protected}/>
           <Route path='/implicit/callback' component={ImplicitCallback} />
         </Security>
-      </Router>
     );
   }
 }
 
-export default App;
+export default withRouter(App);
+
+// src/index.js
+import React from 'react';
+import ReactDOM from 'react-dom';
+import App from './App';
+import { BrowserRouter as Router } from 'react-router-dom';
+
+ReactDOM.render(
+  <Router>
+    <App />
+  </Router>
+, document.getElementById('root'));
 ```
 
 ## Show Login and Logout Buttons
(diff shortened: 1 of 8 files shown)


packages/okta-react/src/Auth.js
```diff
@@ -140,10 +137,16 @@ export default class Auth {
...
     return this._oktaAuth.signOut(options)
       .then(() => {
         if (!options.postLogoutRedirectUri && !this._config.postLogoutRedirectUri) {
-          this._history.push(path || '/');
+          let redirectUri = path || '/';
+          // If a relative path was passed, convert to absolute URI
+          if (redirectUri.charAt(0) === '/') {
+            redirectUri = window.location.origin + redirectUri;
+          }
+          window.location.assign(redirectUri);
```
swiftone: How comfortable with these changes are you?  This rips the history out completely, and I've been operating under the assumption that routers (well, react-router) has that module for a reason.  

I can see that this will end up with a working case for what WE do in our tests, but I don't really know what this will do to user cases where they integrate actual apps that have assumptions and behaviors.

aarongranick-okta: A router allows you to *define* all your "routes" using simple relative paths: /a, /b, etc. The router has options to use a hash instead of path, and to set a [basename](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/BrowserRouter.md#basename-string) which is a path where your app is being served from. So a route "/a" might actually be something like /myapp#/a

As long as you always use the history API, this logic will be applied consistently and transparently and your app doesn't need to worry about hash or basename. Sometimes (like in our case) you want to build a complete URI ahead of time. For these cases, [history.createHref](https://github.com/ReactTraining/history/blob/8bbace1aabbce9fc718a24a2d08c6ac41e9e30b6/modules/createBrowserHistory.js#L150) will apply all the logic using a specific location and output the URI as a string. We are calling this method here, in SecureRoute: https://github.com/okta/okta-oidc-js/pull/680/files#diff-8fd493279f7257b15a10a54019dd087aR21

The Route has a connection to history.location which contains all relevant information needed for history to generate an absolute URI for the current route. When window location is set to this URI, the Router will load the correct route.  This logic *has* to work outside the history api because the user can navigate to any point in the history by using back/forward buttons or the history menu. They can also paste a URL directly into the address bar. In all these cases, the Router watches for changes and transitions to the correct route.

The other logical abstraction React router provides is to run outside of the DOM (because of react-native). We do not support this scenario, our library will only work within a browser (this is true for okta-auth-js, which also uses `window.location.assign()`)

I have high confidence that these changes will cause no issues for users, as long as they are using one of the out-of-the box history objects provided by react-router-dom. There may be a case where someone is using a completely custom history object and using that to intercept the redirect. But I doubt this case actually exists, and it would break with AuthJS 3.0

The logic on `signOut` is changing in 3.0 to perform a postLogoutRedirect automatically, which will either be set by 302 from the login page or by `window.location.assign` by okta-auth-js. So actually this piece of post signout redirect logic will be removed from the SDKs. If someone wants to customize the way signOut redirects they will need to call the new method `closeSession` and then perform the redirect themselves.

packages/okta-react/README.md
```diff
@@ -231,23 +241,23 @@ For PKCE flow, this should be left undefined or set to `['code']`.
...
-}
 
 class App extends Component {
+  customAuthHandler() {
+    // Redirect to the /login page that has a CustomLoginComponent
+    this.props.history.push('/login');
+  }
+
   render() {
+    const customAuthHandler = this.customAuthHandler.bind(this);
```
shuowu-okta: nit: move this line to `constructor`

packages/okta-react/README.md
```diff
@@ -73,28 +73,38 @@ Here is a minimal working example. This example defines 3 routes:
...
           <SecureRoute path='/protected' component={Protected}/>
           <Route path='/implicit/callback' component={ImplicitCallback} />
         </Security>
-      </Router>
     );
   }
 }
 
-export default App;
+export default withRouter(App);
```
swiftone: This HOC is react-router specific.  Hard to genericize this.

PR #692, opened 2020-03-09
fix[okta-react]: Fixes bug where LoginCallback throws error

swiftone:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

When authState in LoginCallback throws, it tries to print an error object inside a p tag which throws a React child error. "Objects are not a valid React child"

Issue Number: N/A

## What is the new behavior?

converts error object to string before rendering in LoginCallback

- [ ] Yes
- [X] No

### DIFF
--- a/packages/okta-react/src/LoginCallback.js
+++ b/packages/okta-react/src/LoginCallback.js
@@ -21,7 +21,7 @@ const LoginCallback = () => {
   }, [authService]);
 
   if(authState.error) { 
-    return <p>${authState.error}</p>;
+    return <p>{authState.error.toString()}</p>;
   }
 
   return null;


packages/okta-react/src/LoginCallback.js
```diff
@@ -21,7 +21,7 @@ const LoginCallback = () => {
   }, [authService]);
 
   if(authState.error) { 
-    return <p>${authState.error}</p>;
+    return <p>{authState.error.toString()}</p>;
```
luispeerez: Shouldn't this print the `authState.error.message` instead of displaying the whole error object?

Screenshot from 2020-03-11 11-54-17

luispeerez: Also, the `authState` docs (https://github.com/okta/okta-oidc-js/tree/master/packages/okta-react#authstate) should be modified, in order to explain that `error` is an object, detailing each property.

luispeerez: Already created a PR with the change and docs modification https://github.com/okta/okta-oidc-js/pull/697 but there's a problem with CI/CD

QuarkNerd: I looked at the last version to decide my change. Version 1.4.1 has this in the  ImplicitCallback component,

```
  componentDidMount() {
    this.props.auth.handleAuthentication()
    .then(() => {
      const location = this.props.auth.getFromUri();
      this.setState({ authenticated: true, location });
    })
    .catch(err => this.setState({ authenticated: false, error: err.toString() }));
  }
```

luispeerez: Yes, you are right, the error is an `AuthSdkError` https://github.com/okta/okta-auth-js/blob/master/packages/okta-auth-js/lib/errors/AuthSdkError.js
And it seems its prototype has the generics `Error` methods, when it calls `toString` it prints the concatenation of the error name and message properties https://developer.mozilla.org/es/docs/Web/JavaScript/Reference/Global_Objects/Error/toString

luispeerez: Hope this get merged soon

swiftone: Me too - it's assigned, it's just not alone.

Confirming the error stringification and adding tests are the requirements to go, but we definitely want to see this merged in.

PR #690, opened 2020-03-06
feat: release 2.0.0

aarongranick-okta:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our [guidelines](/okta/okta-oidc-js/blob/master/CONTRIBUTING.md#commit)
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: 277766

## What is the new behavior?
See CHANGELOG

## Does this PR introduce a breaking change?
- [x] Yes
- [ ] No

## Reviewers
@swiftone 
@robertjd

PR #715, opened 2020-03-23
updates to auth-js 3

swiftone:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: N/A

## What is the new behavior?

- Updates to @okta/okta-auth-js v 3 (which defaults to PKCE)
- Removes wrapping div from `<Security>`

- [x] Yes
- [ ] No

- PKCE is now default. Devs will have to update their Okta Application settings and possibly polyfills for their web apps, or explicitly set `pkce: false`

- `<Security>` no longer provides a wrapping div with customizable `className`.  Devs that depend on that will have to manually provide the div as a child or parent of `<Security>`

### DIFF
--- a/packages/okta-react/test/e2e/harness/src/SessionTokenLogin.js
+++ b/packages/okta-react/test/e2e/harness/src/SessionTokenLogin.js
@@ -29,9 +29,8 @@ export default withOktaAuth(class SessionTokenLogin extends Component {
   }
 
   componentWillMount() {
-    const origin = new URL(this.props.authService._config.issuer).origin;
     this.oktaAuth = new OktaAuth({
-      url: origin
+      issuer: this.props.authService._config.issuer
     });
   }
 
--- a/packages/okta-react/test/jest/authService.test.js
+++ b/packages/okta-react/test/jest/authService.test.js
@@ -820,13 +820,13 @@ describe('AuthService', () => {
       const idToken = {
         idToken: 'fake id'
       };
-      jest.spyOn(authService._oktaAuth.token, 'parseFromUrl').mockReturnValue([accessToken, idToken]);
+      jest.spyOn(authService._oktaAuth.token, 'parseFromUrl').mockReturnValue({ tokens: { accessToken, idToken }});
       jest.spyOn(authService._oktaAuth.tokenManager, 'add');
       return authService.handleAuthentication()
         .then(() => {
           expect(authService._oktaAuth.tokenManager.add).toHaveBeenCalledTimes(2);
-          expect(authService._oktaAuth.tokenManager.add.mock.calls[0]).toEqual(['accessToken', accessToken]);
-          expect(authService._oktaAuth.tokenManager.add.mock.calls[1]).toEqual(['idToken', idToken]);
+          expect(authService._oktaAuth.tokenManager.add.mock.calls[0]).toEqual(['idToken', idToken]);
+          expect(authService._oktaAuth.tokenManager.add.mock.calls[1]).toEqual(['accessToken', accessToken]);
         });
     });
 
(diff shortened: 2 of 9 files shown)


aarongranick-okta:
Just a couple things:
- getUser() - the "token substitution mitigation" logic is now in AuthJS, so it can be removed from this library. See https://github.com/okta/okta-oidc-js/pull/690/files#diff-1c18d8d8315bec14ad494009d7604efaR137
- logOut() - the XHR logout has been moved to a separate method, so we should construct a `postLogoutRedirectUri` from the path argument instead of trying to set the path after logout. See https://github.com/okta/okta-oidc-js/pull/690/files#diff-1c18d8d8315bec14ad494009d7604efaL244

packages/okta-react/test/jest/authService.test.js
```diff
@@ -252,61 +252,46 @@ describe('AuthService', () => {
 
   describe('logout', () => {
 
-    test('by default, it will set history location to "/"', async () => {
+    test('defaults to passing an empty options to signOut', async () => {
       const authService = new AuthService({
         issuer: 'https://foo/oauth2/default',
         client_id: 'foo',
```
aarongranick-okta: FYI: we are no longer accepting these "deprecated" config names with underscores in AngularJS (client_id should be clientId). We should be phasing these out across the board, and remove support in the next major version of React (is this that version?)

packages/okta-react/test/e2e/harness/src/SessionTokenLogin.js
```diff
@@ -29,9 +29,8 @@ export default withOktaAuth(class SessionTokenLogin extends Component {
   }
 
   componentWillMount() {
-    const origin = new URL(this.props.authService._config.issuer).origin;
     this.oktaAuth = new OktaAuth({
-      url: origin
+      issuer: this.props.authService._config.issuer
```
aarongranick-okta: Might make sense to move config to a separate file so it could be shared. Alternatively, you can use `process.env.ISSUER` Even more alternatively, why do we need to create a separate instance of AuthJS here? Why can't we just use `props.authService` ?

packages/okta-react/test/jest/authService.test.js
swiftone: Yes, this is the 3.0 release for React.  I'll update the tests and check to see if we're explicitly converting in the code.

packages/okta-react/test/e2e/harness/src/SessionTokenLogin.js
swiftone: Good question - I'm not sure why this was using a separate instance in the first place.  I'll take a peek.

packages/okta-react/test/jest/authService.test.js
aarongranick-okta: Config is "normalized" (with legacy support) here: https://github.com/okta/okta-oidc-js/pull/715/files#diff-8a51cbb87b5cfd7e092f0410797f00e9R30 We can skip this call if we are ready to deprecate the legacy format.

packages/okta-react/test/e2e/harness/src/SessionTokenLogin.js
swiftone: Looks like this was a separate instance because it's trying to create a different config - just the issuer, not the other params (clientId, redirectUri, onAuthRequired) because this test is explicitly testing the use of the session token.  

Given that this.props.authService._config is set via process.env.ISSUER, I'm inclined to use that rather than introduce env variables to multiple files, unless you're aware of a problem with that.

PR #718, opened 2020-03-26
fix[okta-react]: check is user authenticated default behavior

Valdermeyder:
The current implementation differs from what is mentioned in the documentation "By default, authService will consider a user authenticated if both getIdToken() and getAccessToken() return a value"
Moreover, this leads to the problems when the user doesn't have `idToken` but in the state, the `isAuthenticated` property still has value `true`

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

When the user doesn't have `idToken` but in the state, the `isAuthenticated` property still has the value `true`

Issue Number: N/A

## What is the new behavior?
When the user doesn't have `idToken` or `accessToken` the state `isAuthenticated` property has the value `false`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/okta-react/src/AuthService.js
+++ b/packages/okta-react/src/AuthService.js
@@ -139,7 +139,7 @@ class AuthService {
       const idToken = await this.getIdToken();
 
       // Use external check, or default to isAuthenticated if either the access or id token exist
-      const isAuthenticated = this._config.isAuthenticated ? await this._config.isAuthenticated() : !! ( accessToken || idToken );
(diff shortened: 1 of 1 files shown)


swiftone:
Moved to https://github.com/okta/okta-oidc-js/pull/723 so it can run with our CI tests.  Thanks for the PR @Valdermeyder!

PR #740, opened 2020-04-09
feat[react-native]: add android custom signin support - OKTA-289380

shuowu-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Currenly there is no support in sdk for custom-sigin-in.

## What is the new behavior?
Add custom sign in support for both ios and android. 

**Implementation detail:**
Add one more signIn overloading method in both ios and android bridge code to get the sessionToken from authN client first, then exchange accessToken with sessionToken. 
Promise is returned in newly added signIn method.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
+++ b/packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
@@ -49,6 +49,9 @@ import com.okta.oidc.net.response.UserInfo;
 import com.okta.oidc.storage.SharedPreferenceStorage;
 import com.okta.oidc.util.AuthorizationException;
 
+import java.net.MalformedURLException;
+import java.net.URL;
+
 public class OktaSdkBridgeModule extends ReactContextBaseJavaModule implements ActivityEventListener{
 
     private final ReactApplicationContext reactContext;
@@ -107,8 +110,9 @@ public class OktaSdkBridgeModule extends ReactContextBaseJavaModule implements A
                     .setRequireHardwareBackedKeyStore(requireHardwareBackedKeyStore)
                     .create();
 
+            String authnUrl = getAuthnUrl(discoveryUri);
             this.authnClient = AuthenticationClients.builder()
-                .setOrgUrl(discoveryUri)
+                .setOrgUrl(authnUrl)
                 .build();
 
             promise.resolve(true);
@@ -593,5 +597,10 @@ public class OktaSdkBridgeModule extends ReactContextBaseJavaModule implements A
         } catch (AuthorizationException e) {
             promise.reject(OktaSdkError.OKTA_OIDC_ERROR.getErrorCode(), e.getLocalizedMessage(), e);
         }
+    }
+
+    private String getAuthnUrl(String urlStr) throws MalformedURLException {
+        URL url = new URL(urlStr);
+        return url.getProtocol() + "://" + url.getHost();
     }
 }
--- a/packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
+++ b/packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
@@ -28,6 +28,11 @@ import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.modules.core.RCTNativeAppEventEmitter;
+import com.okta.authn.sdk.AuthenticationException;
+import com.okta.authn.sdk.AuthenticationStateHandlerAdapter;
+import com.okta.authn.sdk.client.AuthenticationClient;
+import com.okta.authn.sdk.client.AuthenticationClients;
+import com.okta.authn.sdk.resource.AuthenticationResponse;
 import com.okta.oidc.AuthorizationStatus;
 import com.okta.oidc.OIDCConfig;
 import com.okta.oidc.Okta;
@@ -50,6 +55,7 @@ public class OktaSdkBridgeModule extends ReactContextBaseJavaModule implements A
     private OIDCConfig config;
     private WebAuthClient webClient;
     private AuthClient authClient;
+    private AuthenticationClient authnClient;
 
     public OktaSdkBridgeModule(ReactApplicationContext reactContext) {
         super(reactContext);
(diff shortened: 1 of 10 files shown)


shuowu:
@IldarAbdullin-okta As discussed in Slack channel, the team is planning to move forwards with the direction in this PR, which is using bridges to communicate with native OIDC SDKs, then expose APIs via javascript.

packages/okta-react-native/android/src/main/java/com/oktareactnative/OktaSdkBridgeModule.java
```diff
@@ -131,12 +145,54 @@ public void signIn() {
...
+        try {
+            authnClient.authenticate(username, password.toCharArray(), null, new AuthenticationStateHandlerAdapter() {
+                @Override
+                public void handleUnknown(AuthenticationResponse authenticationResponse) {
+                    WritableMap params = Arguments.createMap();
+                    params.putString(OktaSdkConstant.ERROR_CODE_KEY, OktaSdkError.SIGN_IN_FAILED.getErrorCode());
+                    params.putString(OktaSdkConstant.ERROR_MSG_KEY, OktaSdkError.SIGN_IN_FAILED.getErrorMessage());
+                    sendEvent(reactContext, OktaSdkConstant.ON_ERROR, params);
+
+                    promise.reject(OktaSdkError.SIGN_IN_FAILED.getErrorCode(), OktaSdkError.SIGN_IN_FAILED.getErrorMessage());
```
FeiChen-okta, line 158: I'm not sure we should be wrapping authn-java with the react-native sdk. The reason is that this code segment does not cover all the cases of authn API. If developers run into issues then they would have to edit this module and add the correct handling. So they would have to modify the SDK itself to fix these issues. This is only used as a sample in the android oidc-sdk because it is in the app layer and developers are responsible for error cases. This puts the responsibility in the sdk level.

FeiChen-okta:
Hi @shuowu I think this will be a major effort if you're choosing to go this route. The reason is that you would have to wrap all functionality of authn-java using react-native. I don't think you would need to do that since you can do the same with authn-js as @IldarAbdullin-okta pointed out. The sample app you can show how to use the authn-js SDK to get a session token and then proceed with sign-in using the OIDC-SDK wrapper that does native sign-in. This would follow the same pattern in Android where the authn API functionality is left up to developers to handle.

IldarAbdullin-okta:
Hey @FeiChen-okta ,

I talked with Robert. The idea for now is to implement primary authentication flow, without handling any extra states in identity pipeline. Looks like primary auth is the most popular use-case for 3rd party RN devs.
Long term solution will be implementing of brand new RN AuthN module in pure js, that will be part `okta-react-native` SDK. It will handle all use cases and states. But for now the proposed solution is the simplest and easiest solution that they can do

FeiChen-okta:
> Hey @FeiChen-okta ,
> 
> I talked with Robert. The idea for now is to implement primary authentication flow, without handling any extra states in identity pipeline. Looks like primary auth is the most popular use-case for 3rd party RN devs.
> Long term solution will be implementing of brand new RN AuthN module in pure js, that will be part `okta-react-native` SDK. It will handle all use cases and states. But for now the proposed solution is the simplest and easiest solution that they can do

@shuowu-okta my concern now is the limitation of authn-java for Android. The issue here is that the authn-java sdk requires Java 8, which is only supported for devices API 24+. We do have a workaround in place https://github.com/okta/okta-oidc-android/tree/authn_android but I'm not sure how this will work using react-native's compilation tools. Android Studio 4.0 will have the ability to use desugar and add Java 8 APIs automatically but that is not released yet. But I don't think react-native tools are using Android Studio. These are some unknowns that will need investigation.

shuowu:
@FeiChen-okta Good catch on the API version! I just did a quick research, looks like the current `MIN_SDK_VERSION` is 19, which means we do need the workaround. Not sure if it's possible to customize depenencies in `build.gradle` based on the api level?

Another option can be acquiring session token via authJS SDK, but JS team have concern about network request may work differently on different platforms.

FeiChen-okta:
> @FeiChen-okta Good catch on the API version! I just did a quick research, looks like the current `MIN_SDK_VERSION` is 19, which means we do need the workaround. Not sure if it's possible to customize depenencies in `build.gradle` based on the api level?
> 
> Another option can be acquiring session token via authJS SDK, but JS team have concern about network request may work differently on different platforms.

@shuowu Gradle doesn't have that ability. You can have separate gradle files and depending on which API level, you copy the correct one over. Not an ideal solution since you have to maintain two gradle files. Or you can use the work around for all versions. I think you should explore if auth-js works. If it does then that will solve the api level issue and might work for iOS also.

shuowu:
@FeiChen-okta I had tried authJS, it works for acquiring session token. 

@aarongranick-okta Do you have any concern if we only use authJs for `session token` purpose? 
I find [issue's with react-native fetch](https://reactnative.dev/docs/network#known-issues-with-fetch-and-cookie-based-authentication), not sure any of the issues related to authJs implementation per `session token` acquisition?
Probably it's also good to add `react-native` field in authJs package.json to limit the usage.

IldarAbdullin-okta:
> https://reactnative.dev/docs/network#known-issues-with-fetch-and-cookie-based-authentication

@shuowu , this issue is not applicable to AuthN. AuthN doesn't use cookies or redirects. AuthN is a set of REST APIs that at the end returns you a session token.

> JS team have concern about network request may work differently on different platforms.

What do you mean here? Do you have any examples?
By my opinion it is more logical to keep business and communication logic in platform independent code rather than move it to native modules

shuowu-okta:
@IldarAbdullin-okta I agree with you. As long as authJs works well for `session token`, and restrict `okta-react-native` only get access to server APIs in authJs , we should be good.

> JS team have concern about network request may work differently on different platforms.

It's the same thing as the shared link.   

@aarongranick-okta @swiftone Do you have any concerns about use authJs to acquire session token instead of leverage native authN modules?

shuowu:
Looks like different SDKs will send different `user-agent` in header. Wondering for `react-native` what `user-agent` should be attached in header?

FeiChen-okta:
> Looks like different SDKs will send different `user-agent` in header. Wondering for `react-native` what `user-agent` should be attached in header?

Not sure. But I would like to know the split for iOS/Android usage. Maybe we can split it by detecting the platform. We have a wiki on the user-agent i'll send it to you.

shuowu-okta:
Close this PR as decision had been made to make changes in `okta-auth-js` to support authN requests for react-native.

PR #751, opened 2020-04-21
feat[react-native]: support primary auth and expose auth client - OKTA-291767

shuowu-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/okta-react-native/CHANGELOG.md
+++ b/packages/okta-react-native/CHANGELOG.md
@@ -1,3 +1,11 @@
+#1.4.0
+
+### Features
+- [`63c0b89`](https://github.com/okta/okta-oidc-js/pull/751/commits/63c0b892ea1bea085ee594aab480e97c41969a32)
+  - Support primary authentication flow
+  - Add `getAuthClient` method to expose `@okta/okta-auth-js` client instance
+  - Add `Promise` support for `authenticate` method
+
 # 1.3.0
 
 ### Features
--- a/packages/okta-react-native/CHANGELOG.md
+++ b/packages/okta-react-native/CHANGELOG.md
@@ -1,7 +1,7 @@
 #1.4.0
 
 ### Features
-- [`63c0b89`](https://github.com/okta/okta-oidc-js/pull/751/commits/63c0b892ea1bea085ee594aab480e97c41969a32)
+- [#751](https://github.com/okta/okta-oidc-js/pull/751)
   - Support primary authentication flow
   - Add `getAuthClient` method to expose `@okta/okta-auth-js` client instance
   - Add `Promise` support for `authenticate` method
--- a/packages/okta-react-native/README.md
+++ b/packages/okta-react-native/README.md
@@ -191,12 +191,36 @@ await createConfig({
 });
 ``` 
 
+### `getAuthClient`
+
+This method will return an instance of [`@okta/okta-auth-js`](https://github.com/okta/okta-auth-js) client to communicate with Okta Authentication API. For more information, please checkout [Okta AuthJs Node JS and React Native Usage](https://github.com/okta/okta-auth-js#node-js-and-react-native-usage) section.
+
 ### `signIn`
 
-This async method will automatically redirect users to your Okta organziation for authentication. It will emit an event once a user successfully signs in. Make sure your event listeners are mounted and unmounted. Note: on iOS there isn't a `onCancelled` event. If the sign in process is cancelled, `onError` will be triggered. 
+This method will handle both `browser-sign-in` and `custom-sign-in` scenarios based on provided options.
+
+This async method will automatically redirect users to your Okta organziation for authentication. It will emit an event once a user successfully signs in. Make sure your event listeners are mounted and unmounted. Note: on iOS there isn't a `onCancelled` event. If the sign in process is cancelled, `onError` will be triggered.
+
+#### `browser-sign-in`
+`browser-sign-in` leverages device's native browser to automatically redirect users to your Okta organziation for authentication. By providing no argument, this method will trigger the `browser-sign-in` flow. It will emit an event once a user successfully signs in. Make sure your event listeners are mounted and unmounted. **Note**: on iOS there isn't a `onCancelled` event. If the sign in process is cancelled, `onError` will be triggered. 
 
 ```javascript
 signIn();
+```
+
+#### `custom-sign-in`
+`custom-sign-in` provides the way to authenticate the user within the native application. By providing `options` object with username and password fields, this method will retrieve `sessionToken` then exchange it for `accessToken`. 
+Both `Promise` and `Event listeners` are supported. This method is leveraging `@okta/okta-auth-js` SDK to perform authentication API request. For more information, please checkout [Okta AuthJs signIn options](https://github.com/okta/okta-auth-js#signinoptions) section.
+
+##### Sample Usage
+```javascript
+signIn({ username: "{username}", password: "{password}" })
+  .then(token => {
+    // consume accessToken from token.access_token
+  })
+  .catch(error => {
+    // handle error
+  })
 ```
 
 ##### Sample Usage
--- a/packages/okta-react-native/README.md
+++ b/packages/okta-react-native/README.md
@@ -10,6 +10,8 @@ This library follows the current best practice for native apps using:
 * [OAuth 2.0 Authorization Code Flow](https://tools.ietf.org/html/rfc6749#section-1.3.1)
 * [Proof Key for Code Exchange (PKCE)](https://tools.ietf.org/html/rfc7636)
 
+This library also exposes APIs to interact with [Authentication API](https://developer.okta.com/docs/api/resources/authn) directly to implement native UI for authentication.
+
 You can learn more on the [Okta + ReactNative](https://developer.okta.com/code/react-native/) page in our documentation. You can also download our [sample application](https://github.com/okta/samples-js-react-native/tree/master/browser-sign-in)
 
 ## Prerequisites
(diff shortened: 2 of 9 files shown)


packages/okta-react-native/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+#1.4.0
+
+### Features
+- [`63c0b89`](https://github.com/okta/okta-oidc-js/pull/751/commits/63c0b892ea1bea085ee594aab480e97c41969a32)
```
aarongranick-okta, line 4: You can just link to the PR (no need for the commit SHA)

aarongranick-okta: Also the link text can be the PR #

packages/okta-react-native/README.md
```diff
@@ -191,14 +191,38 @@ await createConfig({
...
+This method will handle both `browser-sign-in` and `custom-sign-in` scenarios based on provided options.
+
+This async method will automatically redirect users to your Okta organziation for authentication. It will emit an event once a user successfully signs in. Make sure your event listeners are mounted and unmounted. Note: on iOS there isn't a `onCancelled` event. If the sign in process is cancelled, `onError` will be triggered.
+
+#### `browser-sign-in`
+`browser-sign-in` leverages device's native browser to automatically redirect users to your Okta organziation for authentication. By providing no argument, this method will trigger the `browser-sign-in` flow. It will emit an event once a user successfully signs in. Make sure your event listeners are mounted and unmounted. **Note**: on iOS there isn't a `onCancelled` event. If the sign in process is cancelled, `onError` will be triggered.
```
aarongranick-okta, line 205: Is this Note about iOS a bug or a feature?

shuowu-okta: Sort of a bug to declare the inconsistency between ios and android, here I just copied it over.

packages/okta-react-native/CHANGELOG.md
shuowu-okta: Sure, I was just following the current changelog, will change.

packages/okta-react-native/index.js
```diff
@@ -28,6 +33,14 @@ export const createConfig = async({
   assertRedirectUri(redirectUri);
   assertRedirectUri(endSessionRedirectUri);
 
+  const { origin } = Url(discoveryUri);
+  authClient = new OktaAuth({ 
+    issuer: origin,
```
aarongranick-okta, line 38: Using the origin as issuer will be using the ORG level authentication server. Is this what we want?

packages/okta-react-native/index.js
```diff
@@ -49,7 +62,44 @@ export const createConfig = async({
...
+      })
+      .then(token => {
+        if (!token) {
+          throw new Error('Failed to get accessToken');
+        }
+
+        return token;
+      })
+      .catch(error => {
+        throw {
```
aarongranick-okta, line 94: should we be throwing a `new Error` object here? If a bare object is thrown there will be no stack trace.

packages/okta-react-native/index.js
```diff
@@ -70,7 +120,18 @@ export const getIdToken = async() => {
...
 
 export const getUser = async() => {
-  return NativeModules.OktaSdkBridge.getUser();
+  return NativeModules.OktaSdkBridge.getUser()
+    .then(data => {
+      if (typeof data === 'string') {
+        try {
+          return JSON.parse(data);
+        } catch (e) {
+          throw { code: "-600", message: 'Okta Oidc error' };
```
aarongranick-okta, line 129: same note here about throwing bare object, should be `new Error` ?

packages/okta-react-native/index.js
shuowu-okta: As it only makes AuthN API request,  not sure is the origin/baseUrl can work properly here? Or we need to add one more field in the config? Any suggestions?

packages/okta-react-native/CHANGELOG.md
aarongranick-okta: We used to do it the other way with the commits, but it's a hassle to update everytime you rebase. The link to the PR is better for users because they can read our interesting conversations.

packages/okta-react-native/README.md
aarongranick-okta: Is this something we can easily fix? Would be nice to have consistent behavior, as long as we are releasing a new version may as well squeeze some fixes in.

packages/okta-react-native/index.js
aarongranick-okta: Users can configure a custom authorization server it would look like ORIGIN/oauth2/AUTH_SERVER_ID

What is the "discoveryUri"? Is it not the issuer (auth server)?

shuowu-okta: It's for the `.well_known` endpoint. The url may come with `/.well_know/xxx`. Probably expose a new field can be easier.

packages/okta-react-native/README.md
FeiChen-okta: This is a inconsistency between android and iOS. I think in the iOS Bridge you can check if the error is a cancel and then call onCancelled. @IldarAbdullin-okta

shuowu-okta: @IldarAbdullin-okta is the `onCancelled` event currently supported in the ios-oidc SDK?

IldarAbdullin-okta: @shuowu , there is no way programmatically cancel auth flow. If user cancels manually then even will be delivered as error via `onError` callback

IldarAbdullin-okta: Agree with @FeiChen-okta , that probably we can analyze error type in bridge and redirect to `onCancelled` callback if user manually cancelled

packages/okta-react-native/README.md
```diff
@@ -191,14 +191,38 @@ await createConfig({
...
 signIn();
 ```
 
+#### `custom-sign-in`
+`custom-sign-in` provides the way to authenticate the user within the native application. By providing `options` object with username and password fields, this method will retrieve `sessionToken` then exchange it for `accessToken`. 
+Both `Promise` and `Event listeners` are supported. This method is leveraging `@okta/okta-auth-js` SDK to perform authentication API request. For more information, please checkout [Okta AuthJs signIn options](https://github.com/okta/okta-auth-js#signinoptions) section.
+
+##### Sample Usage
+```javascript
+signIn({ username: "{username}", password: "{password}" })
```
IldarAbdullin-okta, line 217: nit: I think we need to re-stucture README, since library now includes AuthN functionality. Probably it can consist of main README that describes 2 type of flows that is possible to handle with the current lib. Main README will contain references to protocol specific READMEs - OIDC and AuthN

shuowu-okta: Added here: https://github.com/okta/okta-oidc-js/pull/751/commits/6588f6247542df78fbc788873c6a398d9b8eef35#diff-86cf020763a482bfcb38ba668c066e5dR13

packages/okta-react-native/ios/ReactNativeOktaSdkBridge.m
```diff
@@ -18,7 +18,12 @@ @interface RCT_EXTERN_MODULE(OktaSdkBridge, RCTEventEmitter)
 
 RCT_EXTERN_METHOD(signIn)
 
-RCT_EXTERN_METHOD(authenticate:(NSString *)sessionToken)
+RCT_EXTERN_METHOD(
```
IldarAbdullin-okta, line 21: Do we really need this breaking change?

packages/okta-react-native/README.md
shuowu-okta: Dug a bit deeper into the bridge and ios SDK's code, right now only throw a general error for cancel action, which is not enough to trigger the onCancelled event listener. Probably will need a bit investigation on ios SDK side.

I would suggest tracking the issue in a separate ticket if it needs to be addressed.

packages/okta-react-native/ios/OktaSdkBridge.swift
```diff
@@ -146,14 +146,18 @@ class OktaSdkBridge: RCTEventEmitter {
...
+    func authenticate(_ sessionToken: String,
+                      promiseResolver: @escaping RCTPromiseResolveBlock,
+                      promiseRejecter: @escaping RCTPromiseRejectBlock) {
         guard let _ = config, let currOktaOidc = oktaOidc else {
             let error = OktaReactNativeError.notConfigured
             let errorDic = [
                 OktaSdkConstant.ERROR_CODE_KEY: error.errorCode,
                 OktaSdkConstant.ERROR_MSG_KEY: error.errorDescription
             ]
             sendEvent(withName: OktaSdkConstant.ON_ERROR, body: errorDic)
```
IldarAbdullin-okta, line 158: If you introduce breaking change with callbacks then do we really need to leave events approach?

packages/okta-react-native/ios/ReactNativeOktaSdkBridge.m
shuowu-okta: As bridge call is always asynced, and this change only introduce return promise to the bridge method, I don't really think it's a breaking change from JS API perspective.

IldarAbdullin-okta: I see, it is just additional capability

packages/okta-react-native/ios/OktaSdkBridge.swift
shuowu-okta: Events can still kept to make sure it's backwards compatible. Also, the events approach would still be needed for the OIDC flow.

packages/okta-react-native/README.md
IldarAbdullin-okta: Yes, separate ticket is fine, as most likely additional work will be required on iOS OIDC SDK side

packages/okta-react-native/ios/OktaSdkBridge.swift
IldarAbdullin-okta: Ok, if it is not a breaking change then we definitely need to keep it

PR #778, opened 2020-05-15
[okta-angular] Release 3.0.0

aarongranick-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: 
OKTA-283293
OKTA-290952

## What is the new behavior?

Exposes the full AuthJS API via class inheritance.
Removes Typescript types for auth-js and configuration-validation
Updates internal test app to NG 7 (for Typescript > 3)
Updates okta-auth-js to 4.0.0 and configuration-validation to 1.0.0

- [x] Yes
- [] No

## Reviewers
@swiftone 
@shuowu


swiftone:
For cumulative changes we should have a PENDING entry in the CHANGELOG, so that nothing gets missed when we push out a version.

(I'm assuming the only reason this isn't getting an immediate release is because you have a few of these small changes going out at once)

packages/okta-angular/src/okta/services/okta.service.ts
```diff
@@ -183,13 +192,13 @@ export class OktaAuthService {
      * Parses the tokens from the callback URL.
      */
     async handleAuthentication(): Promise<void> {
-      const res = await this.oktaAuth.token.parseFromUrl();
+      const res = await this.token.parseFromUrl();
       const tokens = res.tokens;
       if (tokens.accessToken) {
-        this.oktaAuth.tokenManager.add('accessToken', tokens.accessToken as AccessToken);
+        this.tokenManager.add('accessToken', tokens.accessToken as AccessToken);
```
swiftone, line 198: Should we export the token keys from okta-auth-js?    

Shuo's other PR makes these keys configurable, which would break this.

shuowu: Looks like this PR is only target to auth-js@4.0 instead of v4.1.

swiftone: >         "@okta/okta-auth-js": "^4.0.0",

That includes 4.1, right?

My question about exporting the keys stands, either way.

aarongranick-okta: I could modify the semver to `~4.0.0` but there **should be no breaking changes in 4.1**, right?

swiftone: @aarongranick-okta Right.   I'm not saying you should change anything, I'm saying 4.1 should be free of breaking changes.

shuowu (approved):

packages/okta-angular/README.md
```diff
@@ -32,9 +32,10 @@ All of these features are supported by this SDK. Additionally, using this SDK, y
 This library currently supports:
 
 - [OAuth 2.0 Implicit Flow](https://tools.ietf.org/html/rfc6749#section-1.3.2)
-- [OAuth 2.0 Authorization Code Flow](https://tools.ietf.org/html/rfc6749#section-1.3.1) with [Proof Key for Code Exchange (PKCE)](https://tools.ietf.org/html/rfc7636) 
+- [OAuth 2.0 Authorization Code Flow](https://tools.ietf.org/html/rfc6749#section-1.3.1) with [Proof Key for Code Exchange (PKCE)](https://tools.ietf.org/html/rfc7636)
 
 > This library has been tested for compatibility with the following Angular versions: 4, 5, 6, 7, 8, 9
+> :warning: Angular versions older than 7 may not be fully compatible with all dependencies of this library, due to an older Typescript version. You may be able to workaround this issue by setting `skipLibChecks: true` in your `tsconfig.json` file.
```
swiftone, line 38: Do we want to claim compatibility with all of these older versions?

swiftone (approved):

packages/okta-angular/README.md
aarongranick-okta: This is related to use of the "unknown" type within the typings for auth-js, "unknown" was introduced in Typescript 3.0. Angular 6 uses a 2.x version of Typescript so it complains. I don't think this breaks compatibility, but we might get questions about it.

swiftone: I wasn't talking about the details of that particular error, I'm just saying do we want to commit to supporting this set of versions.

aarongranick-okta: I say let's wait until a user has an issue and see if we can update the docs with a simple workaround. If not, we can drop support. I don't think there are that many angular users who fail to upgrade - they should expect problems if using a very old version mixed with current versions of other modules.

PR #826, opened 2020-07-15
fix[react]: staled token state

shuowu:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Currently, the `okta-react` SDK keeps staled authState after user login. 

Issue Number: N/A

## What is the new behavior?
Listen on `expired` event from `okta-auth-js`, then renew (if autoRenew turned on) and update authState accordingly.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/okta-react/src/AuthService.js
+++ b/packages/okta-react/src/AuthService.js
@@ -120,7 +120,6 @@ class AuthService {
   }
 
   async updateAuthState() {
-    console.log('call update auth state');
     // avoid concurrent updates
     if( this._pending.authStateUpdate ) { 
       return this._pending.authStateUpdate.promise;
--- a/packages/okta-react/src/AuthService.js
+++ b/packages/okta-react/src/AuthService.js
@@ -41,7 +41,6 @@ class AuthService {
     this._oktaAuth = new OktaAuth(authConfig);
     this._oktaAuth.userAgent = `${packageInfo.name}/${packageInfo.version} ${this._oktaAuth.userAgent}`;
     this._config = authConfig; // use normalized config
-    this._listeners = {};
     this._pending = {}; // manage overlapping async calls 
 
     this.handleAuthentication = this.handleAuthentication.bind(this);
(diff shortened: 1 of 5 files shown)


shuowu:
@swiftone I found there is already an emitter from `auth-js`, just wrapped that one then exposed from `authService` to reduce our maintenance work. 
Method signatures are still keep unchanged.

packages/okta-react/src/AuthService.js
```diff
@@ -256,17 +257,14 @@ class AuthService {
   }
 
   on( event, callback ) {
-    const subscriberId = this._subscriberCount++;
-    this._listeners[event] = this._listeners[event] || {};
-    this._listeners[event][subscriberId] = callback;
-    return () => { 
-      delete this._listeners[event][subscriberId];
-    }
+    this._oktaAuth.emitter.on(event, callback, this._oktaAuth.emitter);
```
swiftone, line 260: This looks like it lets consumers subscribe to the auth-js events, which is a different thing than the exist okta-react authStateChange event.

We want to push those events to auth-js, but auth-js doesn't yet have the concept of authStateChange, so I'm not clear on how this works.   (Note Security.js uses the authStateChange event)

Is there something I'm missing?

shuowu: Auth-js has similar event emitter as implemented in the react SDK. The change here is just to use the `emitter` exposed by auth-js without introducing a breaking change.

packages/okta-react/src/AuthService.js
```diff
@@ -59,9 +58,11 @@ class AuthService {
     this.emit = this.emit.bind(this);
     this.on = this.on.bind(this);
 
-    this._subscriberCount = 0;
-
     this.clearAuthState();
+
+    // Remove expired token or renew token if autoRenew is true
+    // This process will keep authState synced with states in tokenManager
+    this.on('expired', this.updateAuthState);
```
shuowu, line 65: @swiftone The real fix is here. Right now react sdk only listen on `onSessionExpired` callback, which only handled session expiration but not token expiration. 
auth-js has mechanism to set up an expiration timer after retrieval of a new token, and emit `expired` event when token is about to expire by tracking `expireAt` field from token. 

The change here is to listen on that expired event, then renew token from the tokenManager.get.

packages/okta-react/src/AuthService.js
swiftone: So this still emits the authStateChange event, just using the upstream emitter?   That feels...weird.

You removed the test that checks for the authStateChange event - can we add in a version that checks to see that it can be subscribed to?

PR #848, opened 2020-07-29
fix[react]: onSessionExpired default behaviour

shuowu:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: OKTA-318290

## What is the new behavior?
- Updated `auth-js` dependency to 3.2.2
- Removed default `onSessionExpired` callback
- Added `isPendng` authState when login happens to avoid concurrent login flow
- Moved login logic to `useEffect` in `SecureRoute` component

- [ ] Yes
- [x] No

## Other information
Tested SecureRoute Comp for both okta-hosted-signIn and custom-signIn


packages/okta-react/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 3.0.4
```
swiftone, line 1: Didn't we say this would be a major version bump?

packages/okta-react/README.md
```diff
@@ -314,7 +314,7 @@ These options are used by `Security` to configure the [Auth service][]. The most
 - **onAuthRequired** *(optional)* - callback function. Called when authentication is required. If this is not supplied, `okta-react` redirects to Okta. This callback will receive `authService` as the first function parameter. This is triggered when:
     1. [login](#authserviceloginfromuri-additionalparams) is called
     2. A `SecureRoute` is accessed without authentication
-- **onSessionExpired** *(optional)* - callback function. Called when the Okta SSO session has expired or was ended outside of the application. This SDK adds a default handler which will call [login](#authserviceloginfromuri-additionalparams) to initiate a login flow. Passing a function here will disable the default handler.
+- **onSessionExpired** *(optional)* - callback function. Called when the Okta SSO session has expired or was ended outside of the application. This SDK adds a default handler which will call [clearAuthState](##authserviceclearauthstate) to clear authState in context. Passing a function here will disable the default handler.
```
swiftone, line 317: I thought we had decided that the default handler would do nothing, and we'd separately listen to the expired event to call clearAuthState

packages/okta-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@okta/okta-react",
-  "version": "3.0.3",
+  "version": "3.0.4",
```
swiftone, line 3: I know 3.1.2 is required here, but I thought we were bumping to latest auth-js (3.2.0/3.2.1)

packages/okta-react/README.md
shuowu: I think this change should make sense: https://github.com/okta/okta-oidc-js/pull/848/commits/7311f47c92cf933410f0e7db13164b46d1ef9283

packages/okta-react/package.json
shuowu: I was thinking to roll this fix out first, then update the auth-js latest with https://github.com/okta/okta-auth-js/pull/436 included.
@aarongranick-okta do you have any pending fixes to roll out for `auth-js`?

packages/okta-react/README.md
aarongranick-okta: I believe the effect here is that if the accessToken cannot be renewed, the idToken will also be removed. This *might* be the right logical choice, but there could be a timing error if the idToken is checked first. If we want to invalidate all tokens when accessToken renew fails, we would need to make sure to always try to retrieve accessToken before idToken. This is kind of subtle and error prone. Maybe a better choice is to make default `isAuthenticated` logic only true if BOTH idToken and accessToken are present.

swiftone: That change would definitely be a major version change - discussing that change is what led to the idea of removing isAuthenticated entirely

I also don't think that renewal failure of the accessToken in any way invalidates the idToken.  That requires expiration or revocation.

shuowu-okta: https://github.com/okta/okta-auth-js/blob/master/packages/okta-auth-js/lib/browser/browser.js#L184
Looks like based on current `auth-js` implementation, `onSessionExired` callback will only be triggered on accessToken renew failure, then it would make sense to just clear `accessToken` in context.

But how should we handle `idToken` renew failure and `isAuthenticated` state. I think the purpose for this fix is to update the whole `authState` to keep user out of `SecureRoute`.

swiftone: No - the purpose is to have authState be accurate.  If only accessToken is invalid, then authState is updated.  (updateAuthState instead of clearAuthState).

shuowu-okta: it might will cause infinite loop. error -> updateAuthState -> error -> ...
might need some thoughts on when onSessionExpired be triggered first, imo.

packages/okta-react/src/AuthService.js
```diff
@@ -32,12 +32,6 @@ class AuthService {
     assertClientId(authConfig.clientId);
     assertRedirectUri(authConfig.redirectUri);
 
-    // Automatically enter login flow if session has expired or was ended outside the application
-    // The default behavior can be overriden by passing your own function via config: `config.onSessionExpired`
-    if (!authConfig.onSessionExpired) {
-      authConfig.onSessionExpired = this.login.bind(this);
-    }
```
swiftone, line 39: Is there any behavior we should warn the users not to do to avoid a loop?

packages/okta-react/src/SecureRoute.js
```diff
@@ -20,8 +20,7 @@ const RequireAuth = ({ children }) => {
 
   if(!authState.isAuthenticated) { 
     if(!authState.isPending) { 
-      const fromUri = history.createHref(history.location);
-      authService.login(fromUri);
+      authService.logout();
```
swiftone, line 23: Given the pains we've had in tracking this down as desired behavior, perhaps a comment to explain why not login?

packages/okta-react/test/jest/authService.test.js
```diff
@@ -215,24 +215,23 @@ describe('AuthService', () => {
   });
 
   describe('onSessionExpired', () => {
-    it('By default, sets a handler for "onSessionExpired" which calls login()', () => {
-      jest.spyOn(AuthService.prototype, 'login').mockReturnValue(undefined);
+    it('By default, sets a handler for "onSessionExpired" which calls clearAuthState()', () => {
```
swiftone, line 218: Are these tests still accurate?   I don't see these changes in the PR anymore.

shuowu:
Should also include fixes from https://github.com/okta/okta-auth-js/pull/440 to avoid loop from signOut.

packages/okta-react/src/SecureRoute.js
```diff
@@ -10,19 +10,20 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React from 'react';
+import React, { useEffect } from 'react';
 import { useOktaAuth } from './OktaContext';
-import { useHistory, Route } from 'react-router-dom';
+import { Route } from 'react-router-dom';
```
swiftone, line 15: Where do you use Route?

shuowu: Forgot to move the history logic back, good catch, Thanks!

packages/okta-react/src/SecureRoute.js
```diff
@@ -10,19 +10,20 @@
...
 const RequireAuth = ({ children }) => { 
   const { authService, authState } = useOktaAuth();
-  const history = useHistory();
 
-  if(!authState.isAuthenticated) { 
-    if(!authState.isPending) { 
-      const fromUri = history.createHref(history.location);
-      authService.login(fromUri);
-    }
+  useEffect(() => {
```
swiftone, line 20: I like this use of useEffect

packages/okta-react/src/AuthService.js
swiftone: Bump - I think this is good info to expose to the dev unless it is totally safe.

packages/okta-react/package.json
aarongranick-okta: As we discussed, removing `onSessionExpired` is a breaking change. I also believe we should change the implementation of `isAuthenticated` to assert `accessToken && idToken` to reflect realities of 3rd party cookies and other cases where the app is running - unexpectedly - with only one token. When a token renew fails it will remove the token from storage. React will be listening the event and call updateAuthState. Here we should return false from isAuthenticated if one of the tokens is missing. This will allow a new login flow to occur naturally.

packages/okta-react/src/SecureRoute.js
```diff
@@ -10,19 +10,20 @@
...
+    if(!authState.isAuthenticated && !authState.isPending) { 
+      authService.login();
```
aarongranick-okta, line 22: Is this where it is possible to begin concurrent login attempts? If the authState is updated (due to token changes, etc). Do we need to add aa concurrent check on this or place this call into the PromiseQ?

packages/okta-react/README.md
```diff
@@ -314,7 +314,8 @@ These options are used by `Security` to configure the [Auth service][]. The most
 - **onAuthRequired** *(optional)* - callback function. Called when authentication is required. If this is not supplied, `okta-react` redirects to Okta. This callback will receive `authService` as the first function parameter. This is triggered when:
     1. [login](#authserviceloginfromuri-additionalparams) is called
     2. A `SecureRoute` is accessed without authentication
-- **onSessionExpired** *(optional)* - callback function. Called when the Okta SSO session has expired or was ended outside of the application. This SDK adds a default handler which will call [login](#authserviceloginfromuri-additionalparams) to initiate a login flow. Passing a function here will disable the default handler.
+- **onSessionExpired** *(optional)* - callback function. Called when the Okta SSO session has expired or was ended outside of the application. This SDK provides an empty function as the default behaviour. Passing a function here will disable the default handler.
```
aarongranick-okta, line 317: We have this marked as deprecated in authJS 3.2.2. I think we can remove it from the documentation here.

packages/okta-react/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+
+### Bug Fixes
+
+- [#848](https://github.com/okta/okta-oidc-js/pull/848) Change default `onSessionExpired` behaviour to `authService.clearAuthState`
```
aarongranick-okta, line 5: removes `onSessionExpired` behavior

packages/okta-react/src/SecureRoute.js
shuowu: Yes. Probably the `isPending` state is not enough here as it's triggered by emitting an event, which may still run into concurrent flow.

packages/okta-react/src/AuthService.js
shuowu: Concurrency issue resolved by add pending state to login method. Tested by provide login as customized callback, works with no issue now.

packages/okta-react/src/SecureRoute.js
shuowu: Resolved by providing pending state in the login method to make sure it only be called once.

packages/okta-react/README.md
shuowu: Still kept for this patch fix, marked as deprecated.

packages/okta-react/CHANGELOG.md
swiftone, line 1: I (again) thought we had decided on a major version bump (see comments from Aaron) - what am I missing?

shuowu: I think we decide to make remove default `onSessionExpired` as a patch fix to avoid noise. Then put the `isAuthenticated` behavior change into 4.0.

aarongranick-okta: Right. We agreed that removing `onSessionExpired` is a **fix** not a new feature or breaking change. It was never working as intended, and has mostly negative side effects. The positive side effect was triggering a login if **only** the accessToken is missing, but this can be done better by the user providing an `isAuthenticated` function which asserts `accessToken && idToken`. We will change this to be default logic in the next version but it would likely be considered a breaking change.

packages/okta-react/README.md
```diff
@@ -314,7 +314,9 @@ These options are used by `Security` to configure the [Auth service][]. The most
...
+  > :warning: DO NOT trigger `authService.login()` in this callback. This callback is used inside the `login` method, call it again will trigger the protection logic to end the function.
+- **onSessionExpired (deprecated)** *(optional)* - callback function. Called when the Okta SSO session has expired or was ended outside of the application. This SDK provides an empty function as the default behaviour. Passing a function here will disable the default handler.
+  > :warning: DO NOT trigger token renew process, like `tokenManager.get()` or `tokenManager.renew()`, in this callback as it may end up with infinite loop.
 - **isAuthenticated** *(optional)* - callback function. By default, `authService` will consider a user authenticated if both `getIdToken()` and `getAccessToken()` return a value. Setting a `isAuthenticated` function on the config will skip the default logic and call the supplied function instead. The function should return a Promise and resolve to either true or false.  Note that this is only evaluated when the `auth` code has reason to think the authentication state has changed.  You can call the `authService.updateAuthState()` method to trigger a re-evaluation.
```
aarongranick-okta, line 320: Is this text correct? It says "if both" return a value. But I think the current logic is "if either" return a value. Let's double check this and also indicate that we plan on changing this default logic in the next version. And we recommend the user to provide a function here and assert **both** tokens are present, unless their application can function with only one token present.

packages/okta-react/src/AuthService.js
```diff
@@ -196,12 +190,24 @@ class AuthService {
   }
 
   async login(fromUri, additionalParams) {
+    if(this._pending.handleLogin) {
```
aarongranick-okta, line 193: nice. thanks!

shuowu: the closer the easier to manage, hah

packages/okta-react/CHANGELOG.md
aarongranick-okta: > See comments.
> 
> * major version bump
> * remove onSessionExpired
> * change logic on isAuthenticated to return false if either token is missing
> * double check for concurrency errors with useEffect -> login()

As we discussed we will push breaking change with `isAuthenticated` to subsequent release that includes `okta-auth-js@4.0`.

I think we can update the README to provide better guidance for users of this version. (see comments).

packages/okta-react/src/AuthService.js
shuowu: If login should only be run once, we probably can move the check to auth-js v5.0.

packages/okta-react/README.md
swiftone: We've marked this as deprecated, but don't say anything about what they should be doing instead.

packages/okta-react/src/AuthService.js
```diff
@@ -196,12 +190,24 @@ class AuthService {
...
+    this.emitAuthState({ ...this.getAuthState(), isPending: true });
     // Save the current url before redirect
     this.setFromUri(fromUri); // will save current location if fromUri is undefined
-    if (this._config.onAuthRequired) {
-      return this._config.onAuthRequired(this);
+    try {
+      if (this._config.onAuthRequired) {
+        return await this._config.onAuthRequired(this);
+      }
+      return await this.redirect(additionalParams);
```
swiftone, line 207: Errors are just silently lost - is there anything there we'd want to inform the user of?

packages/okta-react/README.md
```diff
@@ -314,7 +314,9 @@ These options are used by `Security` to configure the [Auth service][]. The most
 - **onAuthRequired** *(optional)* - callback function. Called when authentication is required. If this is not supplied, `okta-react` redirects to Okta. This callback will receive `authService` as the first function parameter. This is triggered when:
     1. [login](#authserviceloginfromuri-additionalparams) is called
     2. A `SecureRoute` is accessed without authentication
-- **onSessionExpired** *(optional)* - callback function. Called when the Okta SSO session has expired or was ended outside of the application. This SDK adds a default handler which will call [login](#authserviceloginfromuri-additionalparams) to initiate a login flow. Passing a function here will disable the default handler.
+  > :warning: DO NOT trigger `authService.login()` in this callback. This callback is used inside the `login` method, call it again will trigger the protection logic to end the function.
```
swiftone, line 317: These warnings are just the advice I wanted, thanks!

packages/okta-react/src/AuthService.js
shuowu: only finally been added here, but without catch, I think users still can catch errors if any happen.

swiftone: *scrunches face up in confusion*
*writes test code*
*blinks furiously*
*writes more test code*
*googles*
Huh.  This is not how I expected finally and errors to interact.  I sit corrected.

packages/okta-react/src/AuthService.js
aarongranick-okta: The underlying `token.getWithRedirect` will be placed in the PromiseQueue, but there is additional logic here. One possible error would be calling `login(myUri)` and then calling `login` again with no parameters. This would overwrite the saved URI and affect the final redirect value.

Another option instead of silently ignoring a concurrent `login` attempt is to actually error out. Maybe we could do this in the next release? What I'd like is a way to compare authState with previous authState so we only trigger login if we are transitioning `isAuthenticated` from pending -> false or true -> false. A transition of false -> false should not start a login flow.

swiftone: > if we are transitioning isAuthenticated from pending -> false or true -> false

Interesting.  I think that level of check is better done in auth-js, but either way it does look like a nice approach to guide developers away from pitfalls that can otherwise _mostly_ work.

packages/okta-react/src/AuthService.js
shuowu: Test added.

shuowu:
> Please double check description of `isAuthenticated` in the README and let us also provide some guidance to verify both tokens. This will cover any side effects of removing `onSessionExpired`

Updated doc, also provided code example.

packages/okta-react/src/AuthService.js
shuowu: `login` might not only be triggered by `isAuthenticated` state change,  users may also call it from callbacks which is out of our control. I think it still makes sense to protect it generally from concurrency cases.

PR #879, opened 2020-08-28
[configuration-validation]: convert to typescript (release 1.0.0)

aarongranick-okta:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: 290951

## What is the new behavior?

Adds typescript declarations
Is an ES module

- [x] Yes
- [] No

Addition of typescript types may interfere with other code which has declared its own version of objects in this library. Those apps should remove their own definitions.


packages/configuration-validation/babel.config.js
```diff
@@ -0,0 +1,10 @@
+const presets = ['@babel/preset-env'];
+const plugins = [];
+
+// Process typescript when running in jest
+if (process.env.NODE_ENV === 'test') {
```
shuowu, line 5: Does the `root` option work here? The test config can be moved into /test folder.

packages/configuration-validation/package.json
```diff
@@ -32,17 +34,38 @@
...
+    "testEnvironment": "jsdom",
+    "testMatch": [
+      "**/test/*.{js,ts}"
+    ],
+    "roots": [
+      "test"
+    ]
   },
   "dependencies": {
+    "@okta/okta-auth-js": "^4.0.0",
```
shuowu, line 68: Looks like it only depends on types from `auth-js`, should it belongs to `dependencies` or `devDependenies`?

shuowu: never mind, looks like the module also exports new types based on the imported types.

packages/configuration-validation/babel.config.js
aarongranick-okta: The tests seem to work fine as .ts files using `babel-jest`. I'm not sure about moving the babel config. I copied this from okta-auth-js. I think the `root` option would apply if we were running the tests from the `/tests` directory (I think it  goes with **working** directory, not **src** directory

PR #884, opened 2020-08-31
fix[react]: store secureRouterReferrerPath in sessionStorage

shuowu:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Adding Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Issue Number: OKTA-282373

## What is the new behavior?
Store `secureRouterReferrerPath` in sessionStorage

- [ ] Yes
- [x] No

### DIFF
--- a/packages/okta-react/src/AuthService.js
+++ b/packages/okta-react/src/AuthService.js
@@ -252,13 +252,13 @@ class AuthService {
     if (fromUri.charAt(0) === '/') {
       fromUri = window.location.origin + fromUri;
     }
-    localStorage.setItem( 'secureRouterReferrerPath', fromUri );
+    sessionStorage.setItem( 'secureRouterReferrerPath', fromUri );
   }
 
   getFromUri() {
     const referrerKey = 'secureRouterReferrerPath';
-    const location = localStorage.getItem(referrerKey) || window.location.origin;
-    localStorage.removeItem(referrerKey);
+    const location = sessionStorage.getItem(referrerKey) || window.location.origin;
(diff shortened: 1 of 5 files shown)


packages/okta-react/src/AuthService.js
```diff
@@ -252,13 +252,13 @@ class AuthService {
     if (fromUri.charAt(0) === '/') {
       fromUri = window.location.origin + fromUri;
     }
-    localStorage.setItem( 'secureRouterReferrerPath', fromUri );
+    sessionStorage.setItem( 'secureRouterReferrerPath', fromUri );
```
swiftone, line 255: This looks like the kind of issue that blew us up when we had conflicting versions of Auth JS between the SIW and outside the SIW.

@aarongranick-okta - as I recall, the solution was to first check for the localStorage version, then fall back to the sessionStorage, but that won't work for this scenario (which is specifically multiple tabs), right?  Am I overthinking this?

shuowu: Not sure, but I saw `okta-angular` is implemented with sessionStorage.

aarongranick-okta: @swiftone I don't think this will affect the widget. Unlike the PKCE meta which is stored by auth flow initializer (SIW) and read by callback handler (SDK). This value is stored by the SDK **before** the redirect to SIW, and read by the SDK **after** the callback has been handled.

